PR #79, opened 2026-04-10
refactor: extract shared proxy utilities into pkg/ (issue #78)

### LINKED ISSUE #78: refactor: extract shared proxy/lifecycle/health/state utilities into pkg/ to eliminate duplication across codex and opencode
*Goal:* Extract duplicated functions from the top-level main packages of databricks-claude, databricks-codex, and databricks-opencode into shared pkg/ packages in databricks-claude. Both consumer repos already import databricks-claude/pkg/ for proxy, childproc, updater, etc. — this extends that pattern to the remaining copy-pasted code.

*Why:* The TLS health-check bug (isProxyHealthy hardcoded http:// while proxyHealthy supported https://) had to be fixed in 3 repos independently. wrapWithLifecycle, watchProxy, listenerPort, saveState, printUpdateNotice, and headlessEnsure are nearly line-for-line identical across all three repos, differing only in the binary name logged and repo-specific config paths. Every bug fix or behavior change requires 3 PRs. Extracting into pkg/ means one fix, one test, one PR.

---

*Duplication inventory (sourcebot-verified)*

Each function below was compared across all three repos. "Identical" means the logic is the same modulo string literals (binary name, log prefix, config path).

*1. proxyHealthy(port int, scheme string) bool*
• Location: main.go in all three repos (claude:952, codex:495, opencode:564)
• Status: *Identical.* 7 lines. TLS InsecureSkipVerify handling included.
• Proposed: pkg/health/health.go → func ProxyHealthy(port int, scheme string) bool

*2. isProxyHealthy(port int) bool*
• Location: hooks.go in all three repos (claude:82, codex:57, opencode:57)
• Status: *Identical.* 7 lines. HTTP-only variant (no TLS).
• Proposed: pkg/health/health.go → func IsProxyHealthy(port int) bool (or just call ProxyHealthy(port, "http"))

*3. watchProxy(port int, handler http.Handler, tlsCert, tlsKey string)*
• Location: main.go in all three repos (claude:969, codex:512, opencode:581)
• Status: *Identical* except log prefix string. 20 lines.
• Proposed: pkg/health/watch.go → func WatchProxy(port int, handler http.Handler, tlsCert, tlsKey, logPrefix string)
• The log prefix becomes a parameter.

*4. wrapWithLifecycle(...) http.Handler*
• Location: main.go in all three repos (claude:476, codex:413, opencode:681)
• Status: claude and codex are *identical* (both use refcount). opencode is *simpler* — no refcount, just idle timeout + /shutdown.
• Proposed: pkg/lifecycle/lifecycle.go → func WrapWithLifecycle(cfg LifecycleConfig) http.Handler
• LifecycleConfig struct includes: Inner http.Handler, RefcountPath string (empty = no refcount), IsOwner bool, IdleTimeout time.Duration, APIKey string, DoneCh chan struct{}, LogPrefix string.
• When RefcountPath is empty, /shutdown always triggers exit (opencode behavior). When set, uses refcount (claude/codex behavior).

*5. listenerPort(ln net.Listener, fallback int) int*
• Location: main.go in claude (572), codex (484), opencode (610)
• Status: *Identical.* 7 lines.
• Proposed: pkg/portbind/portbind.go → func ListenerPort(ln net.Listener, fallback int) int
• portbind package already exists; this is a natural fit.

*6. headlessEnsure(port int)*
• Location: hooks.go in all three repos (claude:19, codex:21, opencode:21)
• Status: *Similar structure.* claude has refcount acquire/release around the logic. codex and opencode do not (rely on idle timeout). All share the same poll-until-healthy loop.
• Proposed: pkg/headless/headless.go → func Ensure(cfg EnsureConfig) where EnsureConfig has Port, ManagedEnvVar, LogPrefix, RefcountPath (optional).
• When RefcountPath is non-empty, acquire/release refcount. When empty, skip (codex/opencode behavior).

*7. headlessRelease(port int)*
• Location: hooks.go in claude only (66). codex/opencode don't have it (no session-end hook).
• Proposed: Keep in databricks-claude main package (not shared). Only one consumer.

*8. printUpdateNotice(cfg updater.Config)*
• Location: main.go in all three repos (claude:855, codex:769, opencode:634)
• Status: *Identical* except log prefix and binary name in the output string.
• Proposed: pkg/updater/updater.go → func PrintUpdateNotice(cfg Config) (add to existing updater package)
• The Config already has BinaryName — use that for the log prefix and output string.

*9. buildUpdaterConfig() updater.Config*
• Location: main.go in all three repos (claude:842, codex:756, opencode:621)
• Status: *NOT extractable.* Each repo passes different RepoSlug, BinaryName, CacheFile. This is 5 lines of repo-specific config. Stays local.

*10. saveState / loadState / persistentState*
• Location: state.go in all three repos
• Status: saveState and loadState are *identical* (atomic write via temp file + rename, JSON marshal/unmarshal). The persistentState struct differs per repo (different fields).
• Proposed: pkg/state/state.go → func Save[T any](path string, v T) error and func Load[T any](path string) (T, error)
• Generic over any JSON-serializable struct. Each repo keeps its own persistentState type and statePath.

*11. resolvePort(portFlag int, state persistentState) int*
• Location: state.go in all three repos
• Status: *Identical logic* (flag > saved > default). But the default port differs (49153, 49154, 49156).
• Proposed: pkg/state/state.go → func ResolvePort(portFlag, savedPort, defaultPort int) int
• Trivial 6-line function, but eliminates the copy.

*12. refcountPathForPort(port int) string*
• Location: main.go in claude (566), codex (478); hooks.go in opencode (68)
• Status: *Same pattern* but different prefix string per repo.
• Proposed: pkg/refcount/refcount.go → func PathForPort(prefix string, port int) string (add to existing refcount pkg)

*13. handleHelp / handlePrintEnv*
• Status: *NOT extractable.* The help text is entirely repo-specific. handlePrintEnv has different parameter lists. These stay local.

---

*What stays local (not extracted)*
• handleHelp — repo-specific help text
• handlePrintEnv — repo-specific parameter lists and output format
• buildUpdaterConfig — 5 lines of repo-specific config literals
• installHooks / uninstallHooks — fundamentally different per repo (claude: settings.json, codex: hooks.json + config.toml feature flag, opencode: JS plugin + opencode.json)
• headlessRelease — only exists in claude
• persistentState struct definition — different fields per repo

---

*New packages and additions to existing packages*

```
pkg/health/          (NEW)
  health.go          — ProxyHealthy(port, scheme), IsProxyHealthy(port)
  watch.go           — WatchProxy(port, handler, tlsCert, tlsKey, logPrefix)
  health_test.go

pkg/lifecycle/       (NEW)
  lifecycle.go       — WrapWithLifecycle(LifecycleConfig) http.Handler
  lifecycle_test.go

pkg/state/           (NEW)
  state.go           — Save[T], Load[T], ResolvePort
  state_test.go

pkg/headless/        (NEW)
  headless.go        — Ensure(EnsureConfig)
  headless_test.go

pkg/updater/         (EXISTING — add function)
  updater.go         — add PrintUpdateNotice(cfg Config)

pkg/portbind/        (EXISTING — add function)
  portbind.go        — add ListenerPort(ln, fallback)

pkg/refcount/        (EXISTING — add function)
  refcount.go        — add PathForPort(prefix, port)
```

---

*Non-regression approach*

1. Each extracted pkg function must have *identical behavior* to the current per-repo implementation. The only allowed changes are parameterizing strings (log prefix, binary name, config path).
2. Every new pkg function gets unit tests. Existing inline tests (if any) move to the pkg test file.
3. databricks-claude updates first: replace local functions with pkg calls, run go test ./... — must pass.
4. databricks-codex and databricks-opencode each get a follow-up PR: bump databricks-claude in go.mod, replace local functions with pkg calls, run go test ./... — must pass before merge.
5. Each consumer PR is reviewed independently. No "big bang" merge.

---

*Implementation order (suggested)*

1. pkg/health (proxyHealthy, isProxyHealthy, watchProxy) — highest value, TLS bug was the catalyst
2. pkg/lifecycle (wrapWithLifecycle) — second highest duplication volume
3. pkg/state (saveState, loadState, resolvePort) — generic helpers
4. pkg/headless (headlessEnsure) — moderate complexity due to optional refcount
5. pkg/updater addition (printUpdateNotice) — trivial
6. pkg/portbind addition (listenerPort) — trivial
7. pkg/refcount addition (PathForPort) — trivial

IceRhymers:
## Summary

Closes #78. Extracts duplicated proxy/lifecycle/health/state utility functions from the top-level main package into shared `pkg/` packages. Both `databricks-codex` and `databricks-opencode` already import `databricks-claude/pkg/` — this extends that pattern to the remaining copy-pasted code.

*New packages:*
• `pkg/health/` — `ProxyHealthy(port, scheme)`, `IsProxyHealthy(port)`, `WatchProxy(port, handler, tlsCert, tlsKey, logPrefix)`
• `pkg/lifecycle/` — `WrapWithLifecycle(Config)` with refcount-optional shutdown and idle timeout
• `pkg/state/` — generic `Save[T]`/`Load[T]` (atomic JSON write) and `ResolvePort(flag, saved, default)`
• `pkg/headless/` — `Ensure(Config)` with optional refcount acquire/release

*Additions to existing packages:*
• `pkg/updater/` — `PrintUpdateNotice(cfg)` using `cfg.BinaryName` for log prefix
• `pkg/portbind/` — `ListenerPort(ln, fallback)`
• `pkg/refcount/` — `PathForPort(prefix, port)`

*Caller wiring:*
• `main.go` and `hooks.go` updated to call the new `pkg/` functions
• Local implementations removed (~185 lines deleted)
• All tests updated to use shared packages

*Behavior:* Identical to current — no logic changes, only parameterization of string literals (log prefix, binary name, config path).

## Test plan

- [x] `go test ./...` passes (all 13 packages)
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [ ] Follow-up PRs for databricks-codex and databricks-opencode to consume the new packages

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/pkg/headless/headless.go
+++ b/pkg/headless/headless.go
@@ -1,0 +1,97 @@
+// Package headless provides the headless-ensure logic shared across
+// databricks-claude, databricks-codex, and databricks-opencode.
+package headless
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/IceRhymers/databricks-claude/pkg/health"
+	"github.com/IceRhymers/databricks-claude/pkg/refcount"
+)
+
+// Config holds the parameters for Ensure.
+type Config struct {
+	// Port is the proxy port to check/start.
+	Port int
+
+	// Scheme is "http" or "https" for health checks. Typically "http" for
+	// the headless-ensure path since TLS is not yet negotiated.
+	Scheme string
+
+	// ManagedEnvVar is the environment variable that, when set to "1",
+	// causes Ensure to skip (e.g. "DATABRICKS_CLAUDE_MANAGED").
+	ManagedEnvVar string
+
+	// BinaryPath is the path to the binary to launch in headless mode.
+	// If empty, os.Executable() is used.
+	BinaryPath string
+
+	// LogPrefix is the string used in log messages (e.g. "databricks-claude").
+	LogPrefix string
+
+	// RefcountPath, when non-empty, causes Ensure to acquire/release the
+	// refcount around the operation (claude behavior). When empty, refcount
+	// is skipped (codex/opencode behavior).
+	RefcountPath string
+}
+
+// Ensure checks whether the proxy is healthy on the given port.
+// If not, it starts a detached headless proxy and polls until ready (max 10s).
+func Ensure(cfg Config) {
+	if cfg.ManagedEnvVar != "" && os.Getenv(cfg.ManagedEnvVar) == "1" {
+		log.Printf("%s: --headless-ensure: skipped (managed session)", cfg.LogPrefix)
+		return
+	}
+
+	// Acquire refcount FIRST so every ensure/release pair is symmetric.
+	if cfg.RefcountPath != "" {
+		if err := refcount.Acquire(cfg.RefcountPath); err != nil {
+			log.Printf("%s: --headless-ensure: refcount acquire warning: %v", cfg.LogPrefix, err)
+		}
+	}
+
+	if health.IsProxyHealthy(cfg.Port) {
+		return // already running, refcount incremented
+	}
+
+	self := cfg.BinaryPath
+	if self == "" {
+		var err error
+		self, err = os.Executable()
+		if err != nil {
+			if cfg.RefcountPath != "" {
+				refcount.Release(cfg.RefcountPath) // undo acquire on failure
+			}
+			log.Fatalf("%s: --headless-ensure: cannot find self: %v", cfg.LogPrefix, err)
+		}
+	}
+
+	cmd := exec.Command(self, "--headless", fmt.Sprintf("--port=%d", cfg.Port))
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+	if err := cmd.Start(); err != nil {
+		if cfg.RefcountPath != "" {
+			refcount.Release(cfg.RefcountPath) // undo acquire on failure
+		}
+		log.Fatalf("%s: --headless-ensure: failed to start proxy: %v", cfg.LogPrefix, err)
+	}
+	if err := cmd.Process.Release(); err != nil {
+		log.Printf("%s: --headless-ensure: release warning: %v", cfg.LogPrefix, err)
+	}
+
+	// Poll until healthy or timeout.
+	for i := 0; i < 20; i++ {
+		time.Sleep(500 * time.Millisecond)
+		if health.IsProxyHealthy(cfg.Port) {
+			return
+		}
+	}
+	if cfg.RefcountPath != "" {
+		refcount.Release(cfg.RefcountPath) // undo acquire on failure
+	}
+	log.Fatalf("%s: --headless-ensure: proxy did not become healthy within 10s", cfg.LogPrefix)
+}
--- a/hooks.go
+++ b/hooks.go
@@ -6,10 +6,10 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"time"
 
+	"github.com/IceRhymers/databricks-claude/pkg/headless"
 	"github.com/IceRhymers/databricks-claude/pkg/refcount"
 )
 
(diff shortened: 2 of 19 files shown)


pkg/headless/headless.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := 0; i < 20; i++ {
+		time.Sleep(500 * time.Millisecond)
+		if health.IsProxyHealthy(cfg.Port) {
+			return
+		}
+	}
+	if cfg.RefcountPath != "" {
+		refcount.Release(cfg.RefcountPath) // undo acquire on failure
+	}
+	log.Fatalf("%s: --headless-ensure: proxy did not become healthy within 10s", cfg.LogPrefix)
```
IceRhymers, line 96: #claw why are we putting headless-ensure here in the shared package but not release? I know that claude can only use the release. I would opt to not include access to it in the ohter cli's surface. But their functionality is in the same shared domain. But im open to other arguments.

IceRhymers:
The split is because the two operations cross different process boundaries.

`headless.Ensure` is *client-side*: it runs in the Claude Code hook process, touches the refcount file directly, then launches/polls the proxy. Everything happens before the proxy is alive.

`headlessRelease` is also *client-side* but its mechanism is different — it sends `POST /shutdown` over HTTP to the already-running proxy process. The proxy handles the refcount decrement internally (via `pkg/lifecycle`). The client doesn't touch the refcount file at all on the way out.

So if we added `Release(cfg Config)` to `pkg/headless`, it would need to know about HTTP and the shutdown endpoint — making headless responsible for both a filesystem concern (acquire) and a network concern (HTTP shutdown). That's two different interfaces to the proxy lifecycle, not one.

Your instinct that they're the same domain is right. I'd call the real home `pkg/lifecycle` (which owns the server-side shutdown + refcount decrement), not `pkg/headless`. The client-side hooks.go wiring is thin enough that it doesn't need a shared abstraction.

If it's bugging you we could rename `pkg/headless` → `pkg/ensure` to make the one-directional scope explicit. Happy to do that before merge.

🦀 *(reply via claw)*

PR #84, opened 2026-04-10
fix: scheme-aware health check and TLS arg forwarding in headless.Ensure

### LINKED ISSUE #83: fix: pkg/headless TLS health-check and subprocess arg forwarding
## Summary

Two critical bugs in `pkg/headless/headless.go` that will break TLS-configured users when databricks-codex and databricks-opencode migrate to use `headless.Ensure`.

**This is a prerequisite for the codex/opencode pkg/ extraction refactor.**

## Bug 1: HTTP-only health check

Lines 57 and 89 call `health.IsProxyHealthy(cfg.Port)` which always uses HTTP. When a proxy is configured with TLS (HTTPS), this health check will always return false, causing `headless.Ensure` to repeatedly try to start a new proxy until the timeout.

**Fix:** Replace `health.IsProxyHealthy(cfg.Port)` with `health.ProxyHealthy(cfg.Port, cfg.Scheme)` at both locations.

Note: after this fix, `health.IsProxyHealthy` may be dead code — consider removing it.

## Bug 2: TLS args not forwarded to subprocess

When `headless.Ensure` spawns a detached proxy process, it does not pass `--tls-cert`/`--tls-key` flags. The local implementations in both databricks-codex and databricks-opencode do pass these flags.

**Fix:** Add `TLSCert string` and `TLSKey string` fields to `headless.Config`. When non-empty, append `--tls-cert=<cert>` and `--tls-key=<key>` to the subprocess args.

## Callers after fix

When calling `headless.Ensure`, callers must:
- Set `cfg.Scheme` based on whether TLS is configured (`"https"` if cert/key present, `"http"` otherwise)
- Set `cfg.TLSCert` and `cfg.TLSKey` from saved state

## Acceptance Criteria

- [ ] `health.ProxyHealthy(cfg.Port, cfg.Scheme)` used at both health-check sites
- [ ] `headless.Config` has `TLSCert` and `TLSKey` string fields
- [ ] Subprocess receives `--tls-cert`/`--tls-key` args when non-empty
- [ ] `go test ./...` passes
- [ ] Verified: TLS-configured proxy is correctly detected as healthy by `headless.Ensure`

IceRhymers:
## Summary

Closes #83

- *Bug 1*: `headless.Ensure` called `health.IsProxyHealthy` (HTTP-only) instead of `health.ProxyHealthy(port, scheme)`. When the proxy runs with TLS, the health check always failed. Now uses the scheme-aware variant.
- *Bug 2*: When spawning a detached proxy subprocess, `--tls-cert` and `--tls-key` flags were not forwarded. Added `TLSCert` and `TLSKey` fields to `headless.Config` and appended them to subprocess args when set.
- Removed dead code: `IsProxyHealthy` from `pkg/health` (no remaining callers).

## Test plan

- [x] `TestBuildArgs_NoTLS` — TLS args absent when TLSCert/TLSKey empty
- [x] `TestBuildArgs_WithTLSCert` — --tls-cert appears when TLSCert set
- [x] `TestBuildArgs_WithTLSKey` — --tls-key appears when TLSKey set
- [x] `TestBuildArgs_WithTLSCertAndKey` — both flags appear together
- [x] `TestProxyHealthy_SchemeMismatch` — HTTP server not detected as healthy with https scheme
- [x] `TestEnsure_AlreadyHealthy` — updated to pass Scheme field
- [x] `go build ./...` passes
- [x] `go vet ./...` passes

### DIFF
--- a/pkg/headless/headless.go
+++ b/pkg/headless/headless.go
@@ -37,6 +37,14 @@ type Config struct {
 	// refcount around the operation (claude behavior). When empty, refcount
 	// is skipped (codex/opencode behavior).
 	RefcountPath string
+
+	// TLSCert is the path to the TLS certificate file.
+	// When non-empty, --tls-cert is forwarded to the subprocess.
+	TLSCert string
+
+	// TLSKey is the path to the TLS key file.
+	// When non-empty, --tls-key is forwarded to the subprocess.
+	TLSKey string
 }
 
 // Ensure checks whether the proxy is healthy on the given port.
@@ -54,7 +62,7 @@ func Ensure(cfg Config) {
 		}
 	}
 
-	if health.IsProxyHealthy(cfg.Port) {
+	if health.ProxyHealthy(cfg.Port, cfg.Scheme) {
 		return // already running, refcount incremented
 	}
 
@@ -70,7 +78,7 @@ func Ensure(cfg Config) {
 		}
 	}
 
-	cmd := exec.Command(self, "--headless", fmt.Sprintf("--port=%d", cfg.Port))
+	cmd := exec.Command(self, buildArgs(cfg)...)
 	cmd.Stdout = nil
 	cmd.Stderr = nil
 	if err := cmd.Start(); err != nil {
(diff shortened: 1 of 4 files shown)


IceRhymers:
#claw ci failed tackle this with a ralplan and an oopus agent and try again

IceRhymers:
Root package failure was caused by `headlessEnsure` in `hooks.go` not setting `Scheme`, so `health.ProxyHealthy(port, "")` produced an invalid URL.

Fixed: `Ensure` now defaults `Scheme` to `"http"` when empty — callers that don't set it (like the existing claude hooks.go) get the correct behavior, while callers that do set it (TLS configs) get scheme-aware checks.

Pushed to same branch.

🦀 *(reply via claw)*

PR #93, opened 2026-04-27
feat: --credential-helper and --generate-desktop-config for Claude Desktop

### LINKED ISSUE #92: feat: generate MDM config for Claude Desktop inference endpoint
## Summary

Add \`--credential-helper\` and \`--generate-desktop-config\` flags to \`databricks-claude\` for full Claude Desktop inference integration — no static token, no UI navigation.

**All MDM keys now confirmed from actual Desktop exports.** See files below.

---

## Confirmed MDM key names

From Desktop's exported config (proxy variant — credential helper overrides \`inferenceGatewayApiKey\`):

| Key | Type | Value |
|-----|------|-------|
| \`disableDeploymentModeChooser\` | bool | \`true\` — hides provider switcher in UI |
| \`inferenceProvider\` | string | \`"gateway"\` |
| \`inferenceGatewayBaseUrl\` | string | AI Gateway URL (or \`http://127.0.0.1:49153\` for proxy path) |
| \`inferenceGatewayApiKey\` | string | Placeholder required (e.g. \`"managed-by-credential-helper"\`) — helper overrides it |
| \`inferenceGatewayAuthScheme\` | string | \`"bearer"\` — Desktop adds \`Bearer \` prefix to the token |
| \`inferenceModels\` | JSON string | Array of \`{"name": "...", "supports1m": true}\` objects |
| \`inferenceCredentialHelper\` | string | Absolute path to the \`databricks-claude\` binary |
| \`inferenceCredentialHelperTtlSec\` | int | \`55\` (Databricks tokens are 1hr; refresh before expiry) |
| \`isClaudeCodeForDesktopEnabled\` | bool | \`true\` |
| \`isDesktopExtensionEnabled\` | bool | \`true\` |
| \`isDesktopExtensionDirectoryEnabled\` | bool | \`true\` |
| \`isDesktopExtensionSignatureRequired\` | bool | \`false\` |
| \`isLocalDevMcpEnabled\` | bool | \`true\` |

**macOS plist note:** \`inferenceModels\` must be a JSON-encoded **string** in the plist, not a native array.

**Windows reg note:** \`disableDeploymentModeChooser\` is \`dword:00000001\`; all other values are \`REG_SZ\`.

---

## Confirmed model list

\`\`\`json
[
  {"name": "databricks-claude-opus-4-7", "supports1m": true},
  {"name": "databricks-claude-opus-4-6", "supports1m": true},
  {"name": "databricks-claude-sonnet-4-6", "supports1m": true},
  {"name": "databricks-claude-sonnet-4-5", "supports1m": true},
  {"name": "databricks-claude-haiku-4-5"}
]
\`\`\`

---

### \`--credential-helper\` flag

Called by Desktop with no arguments. Outputs a fresh Databricks OAuth token to stdout (trimmed), exits 0. Since \`inferenceGatewayAuthScheme: "bearer"\` is set, Desktop handles the \`Bearer \` prefix — output is just the raw token:

\`\`\`
dapixxx...
\`\`\`

Or JSON form (also valid for gateway):
\`\`\`json
{"token": "dapixxx..."}
\`\`\`

Nothing on stderr. Uses existing token-fetch logic from \`token.go\`.

### \`--generate-desktop-config\` flag

Auto-discovers the Databricks AI Gateway URL (same logic as existing auto-discovery), gets the absolute path of the current binary via \`os.Executable()\`, then writes:
- macOS → \`databricks-claude-desktop.mobileconfig\`
- Windows → \`databricks-claude-desktop.reg\`

The \`inferenceGatewayApiKey\` is set to \`"managed-by-credential-helper"\` (placeholder — credential helper overrides it at runtime).

---

## Acceptance criteria

- [ ] \`--credential-helper\` outputs a fresh Databricks OAuth token to stdout, exits 0, nothing on stderr
- [ ] Desktop re-invokes helper on TTL expiry transparently (no user prompt)
- [ ] \`--generate-desktop-config\` writes valid \`.mobileconfig\` (macOS) with correct plist encoding for \`inferenceModels\`
- [ ] \`--generate-desktop-config\` writes valid \`.reg\` (Windows) with correct types (\`dword\` for bool, \`REG_SZ\` for strings)
- [ ] Generated config uses \`os.Executable()\` for the credential helper path (not hardcoded)
- [ ] Full end-to-end inference call through Desktop succeeds with a real Databricks workspace
- [ ] Works on macOS and Windows

## Out of scope

- Pushing configs via actual MDM systems (Jamf, Intune) — file generation only
- Linux

IceRhymers:
## Summary

Adds two new flags to the `databricks-claude` binary so it can act as the inference provider for Claude Desktop via the new MDM configuration scheme.

- `--credential-helper` — invoked by Claude Desktop on demand for `inferenceCredentialHelper`. Resolves the active profile (`--profile` flag > saved state file > `DEFAULT`), fetches a fresh Databricks OAuth token using the existing `token.go` logic, and writes the raw token to stdout. Stays silent on stderr on success and exits non-zero on failure.
- `--generate-desktop-config` — discovers the AI Gateway URL from the active profile (re-using `DiscoverHost` + `ConstructGatewayURL`), captures the absolute path of the running binary via `os.Executable()`, and writes a platform-specific MDM config:
  - **macOS** → `databricks-claude-desktop.mobileconfig` (Apple Configuration Profile XML plist with two distinct UUIDs, generated from `crypto/rand`).
  - **Windows** → `databricks-claude-desktop.reg` (registry script under `HKCU\SOFTWARE\Policies\Claude` with CRLF line endings, escaped backslashes, and escaped JSON quotes for `inferenceModels`).
  - On other OSes both files are written so the user can transfer them. `--output <path>` overrides the default filename and infers the format from the extension.
- Help text and shell completions updated; the new paths are handled before `parseArgs` runs so the existing return tuple and its tests are untouched.
- Pure stdlib only — no new dependencies introduced.

## Test plan

- [x] `go test ./... -count=1` (all existing + new tests pass)
- [x] `go vet ./...` clean
- [x] `go build .` succeeds; `--help` lists the new flags
- [x] Unit tests cover: UUID format/uniqueness, mobileconfig content/escaping/UUID separation, .reg content/escaping/CRLF, `regEscape`, `plistEscape`, `extractProfileFlag`, `extractOutputFlag`, `hasFlag`
- [x] Manual smoke test on macOS: run `databricks-claude --generate-desktop-config`, install the resulting `.mobileconfig`, restart Claude Desktop, verify it calls back into `--credential-helper` and gets a working token
- [ ] Manual smoke test on Windows: import the resulting `.reg`, restart Claude Desktop, verify the same end-to-end flow

Closes #92

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,7 +69,13 @@ func main() {
 	// --credential-helper — invoked by Claude Desktop (inferenceCredentialHelper).
 	// Must be handled VERY early: outputs only the raw token to stdout, exits.
 	// Profile resolution mirrors the main flow (state file > "DEFAULT").
-	if hasFlag(os.Args[1:], "--credential-helper") {
+	//
+	// Two dispatch paths:
+	//   1. Explicit `--credential-helper` flag — useful for scripting/debug.
+	//   2. argv[0] alias `databricks-claude-credential-helper` — Desktop's
+	//      mobileconfig accepts only a path, no args; install methods drop a
+	//      symlink at this name pointing at the main binary.
+	if hasFlag(os.Args[1:], "--credential-helper") || isCredentialHelperBinaryName(os.Args[0]) {
 		runCredentialHelper(extractProfileFlag(os.Args[1:]))
 		// runCredentialHelper always calls os.Exit; this return is unreachable.
 		return
@@ -82,6 +88,7 @@ func main() {
 		runGenerateDesktopConfig(
 			extractProfileFlag(os.Args[1:]),
 			extractOutputFlag(os.Args[1:]),
+			extractBinaryPathFlag(os.Args[1:]),
 		)
 		return
 	}
@@ -737,11 +744,16 @@ Databricks-Claude Flags:
   --install-hooks              Install SessionStart/Stop hooks into ~/.claude/settings.json
   --uninstall-hooks            Remove databricks-claude hooks from ~/.claude/settings.json
   --no-update-check            Skip the automatic update check on startup
-  --credential-helper          Print a fresh Databricks token to stdout (called by Claude
-                               Desktop's inferenceCredentialHelper); honours --profile
+  --credential-helper          Print a fresh Databricks token to stdout (also dispatched
+                               via the databricks-claude-credential-helper symlink that
+                               Claude Desktop's inferenceCredentialHelper points at);
+                               honours --profile
   --generate-desktop-config    Write a Claude Desktop MDM config (.mobileconfig on macOS,
-                               .reg on Windows); honours --profile and --output
+                               .reg on Windows); honours --profile, --output, --binary-path
   --output string              Explicit output path for --generate-desktop-config
+  --binary-path string         Credential-helper path to embed in the generated config
+                               (default: derived from the running binary). Use this for
+                               MDM rollouts so one config works on every endpoint.
   --version                    Print version and exit
   --help, -h                   Show this help message
 
--- a/main.go
+++ b/main.go
@@ -66,31 +66,20 @@ func main() {
 		os.Exit(0)
 	}
 
-	// --credential-helper — invoked by Claude Desktop (inferenceCredentialHelper).
-	// Must be handled VERY early: outputs only the raw token to stdout, exits.
-	// Profile resolution mirrors the main flow (state file > "DEFAULT").
-	//
-	// Two dispatch paths:
-	//   1. Explicit `--credential-helper` flag — useful for scripting/debug.
-	//   2. argv[0] alias `databricks-claude-credential-helper` — Desktop's
-	//      mobileconfig accepts only a path, no args; install methods drop a
-	//      symlink at this name pointing at the main binary.
-	if hasFlag(os.Args[1:], "--credential-helper") || isCredentialHelperBinaryName(os.Args[0]) {
+	// argv[0] alias `databricks-claude-credential-helper` — Desktop's
+	// mobileconfig accepts only a path with no arguments; install methods
+	// drop a symlink at this name pointing at the main binary so Desktop's
+	// inferenceCredentialHelper can target a stable path.
+	if isCredentialHelperBinaryName(os.Args[0]) {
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -717,6 +744,16 @@ Databricks-Claude Flags:
   --install-hooks              Install SessionStart/Stop hooks into ~/.claude/settings.json
   --uninstall-hooks            Remove databricks-claude hooks from ~/.claude/settings.json
   --no-update-check            Skip the automatic update check on startup
+  --credential-helper          Print a fresh Databricks token to stdout (also dispatched
+                               via the databricks-claude-credential-helper symlink that
+                               Claude Desktop's inferenceCredentialHelper points at);
+                               honours --profile
+  --generate-desktop-config    Write a Claude Desktop MDM config (.mobileconfig on macOS,
```
IceRhymers: this is currently generating a the file based on the detected OS, but wouldn't it be better to allow the generation of both artifacts?

token.go
```diff
@@ -6,13 +6,76 @@ import (
...
+// resolveDatabricksCLI returns an executable path for the Databricks CLI.
+// Lookup order:
+//  1. Absolute or path-qualified cmdName → returned unchanged (back-compat for tests).
+//  2. $DATABRICKS_CLI env override, if set and executable.
+//  3. exec.LookPath(cmdName), which honors the inherited PATH.
+//  4. A scan of common install dirs (/usr/local/bin, /opt/homebrew/bin, ~/.local/bin, ~/go/bin, ~/bin).
+//
+// If none match, cmdName is returned unchanged so the eventual exec error
+// surfaces with its original message.
+func resolveDatabricksCLI(cmdName string) string {
```
IceRhymers, line 40: Wouldn't someone with an MDM also want to point to the databricks-cli-path installed like they would with this binary? Then they can pin the path. But we'd need to get it into the CLI somehow over the symlink. Push back if searching is still the best path forward.

IceRhymers (commented):
No provided documentation of the new flags in the README either.

PR #134, opened 2026-05-05
chore: release 0.17.0

IceRhymers:
### Refactors
- \`parseArgs\` 31-value tuple replaced with an \`Args\` struct — all flag values now accessed as \`a.FieldName\` throughout \`main.go\`

### Bug Fixes
- **Stale proxy cleanup**: on startup, \`ensureConfig\` now probes existing \`127.0.0.1:<port>\` entries with a 200ms TCP dial and removes dead ones before writing the new URL — fixes silent \`connection refused\` after crash/reinstall
- **Library code safety**: \`pkg/proxy.NewServer\` and \`pkg/headless.Ensure\` now return errors instead of calling \`log.Fatalf\`; \`authcheck\` resolves the Databricks CLI binary via fallback dirs (fixes silent false-negative on GUI-launched sessions where bare \`databricks\` is not on PATH)
- **Secret sanitization**: adds case-insensitive Bearer, Basic auth, \`access_token\` JSON field, and \`DATABRICKS_TOKEN\` env-var patterns; \`--idle-timeout\` now rejects bare integers with a clear error message
- **Concurrent-write safety**: \`os.CreateTemp\` used for unique tmp filenames in settings writers — eliminates corruption when two processes start simultaneously
- **Proxy reliability**: skips \`Authorization\` header injection when token is empty; fixes WebSocket goroutine leak (single \`<-done\` replaced with explicit conn close + second receive)
- **Lifecycle**: properly wraps handler on health-watcher takeover so \`/shutdown\` triggers a clean shutdown after promotion

### Improvements
- \`strings.HasPrefix\` replaces hand-rolled slice-bounds prefix checks; adds \`https://\` variant for localhost suppression
- README aligned with current CLI flags and \`--print-env\` output


IceRhymers:
#claw look at the diff from this and the latest release to have a better and more updated changelog. also this should be version 0.17.0 since we had a refactor.

IceRhymers:
Updated to **0.17.0** (minor bump for the `parseArgs` struct refactor) and rewrote the changelog:

- Deduplicated the stale-proxy entry (was listed twice from merge + squash commit)
- Dropped the internal `isExecutableFile` shim entry (CI fix, not user-facing)  
- Added a **Refactors** section for the `Args` struct change
- Expanded each bug fix description with the actual behavior change
- Added an **Improvements** section for the HasPrefix cleanup and README alignment

🦀 *(reply via claw)*

PR #137, opened 2026-05-06
fix: return errors from NewServer/Ensure; resolve CLI path in authcheck

### LINKED ISSUE #120: fix: log.Fatalf in library code and hardcoded CLI path in authcheck crash the host process
## Problem

Two related reliability issues that can crash the host Claude process or silently fail auth:

### 1. `authcheck.IsAuthenticated` hardcodes `"databricks"` as the CLI name

`pkg/authcheck/authcheck.go` lines 18-26:

```go
// IsAuthenticated returns true if a valid token can be fetched for the given
// Databricks profile without triggering an interactive login.
func IsAuthenticated(profile string) bool {
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()
	out, err := execCommandContext(ctx, "databricks", "auth", "token", "--profile", profile).Output()
	if err != nil {
		return false
	}
	return strings.Contains(string(out), "access_token")
}
```

`EnsureAuthenticated` (same file, line 37) has the same problem:

```go
cmd := execCommand("databricks", "auth", "login", "--profile", profile)
```

Meanwhile `token.go` lines 17-28 explicitly warns:

```go
// fallbackCLIDirs lists install locations to probe when "databricks" is not on
// PATH. Order matters: most-likely first. GUI-launched subprocesses (e.g.
// Claude Desktop invoking the credential helper) inherit launchd's minimal
// PATH (/usr/bin:/bin:/usr/sbin:/sbin), which omits all of these.
var fallbackCLIDirs = []string{
	"/usr/local/bin",
	"/opt/homebrew/bin",
	"/opt/homebrew/sbin",
	".local/bin", // resolved against $HOME
	"go/bin",     // resolved against $HOME
	"bin",        // resolved against $HOME
}
```

and the token fetcher correctly resolves the CLI before invoking it (`token.go` line 97):

```go
cmd := exec.CommandContext(fetchCtx, resolveDatabricksCLI(f.cmdName), "auth", "token", "--profile", f.profile)
```

So on a GUI launch, `IsAuthenticated` silently returns `false` (because bare `databricks` isn't on the inherited PATH), while `token.go` would still work because it uses the resolved path. The user sees a misleading "not authenticated" state. `EnsureAuthenticated` then tries to run `databricks auth login` with the same bare name, which fails for the same reason.

### 2. `log.Fatalf` inside exported library functions kills the host process

`log.Fatalf` calls `os.Exit(1)` unconditionally. In library code (exported functions and hook callbacks reused across binaries), this terminates the entire host Claude process — not just the proxy or the hook.

**`pkg/proxy/proxy.go`** — inside the exported `NewServer(config *Config) http.Handler` (line 277):

```go
inferenceUpstream, err := url.Parse(config.InferenceUpstream)
if err != nil {
	log.Fatalf("databricks-claude: invalid InferenceUpstream %q: %v", config.InferenceUpstream, err)  // line 282
}

otelUpstream, err := url.Parse(config.OTELUpstream)
if err != nil {
	log.Fatalf("databricks-claude: invalid OTELUpstream %q: %v", config.OTELUpstream, err)  // line 287
}
```

A misconfigured upstream URL takes down the whole process instead of returning an error a caller could surface or recover from.

**`pkg/headless/headless.go`** — inside the exported `Ensure(cfg Config)` (the SessionStart-driven entrypoint, line 52):

```go
self, err = os.Executable()
if err != nil {
	if cfg.RefcountPath != "" {
		refcount.Release(cfg.RefcountPath)
	}
	log.Fatalf("%s: --headless-ensure: cannot find self: %v", cfg.LogPrefix, err)  // line 80
}
...
if err := cmd.Start(); err != nil {
	if cfg.RefcountPath != "" {
		refcount.Release(cfg.RefcountPath)
	}
	log.Fatalf("%s: --headless-ensure: failed to start proxy: %v", cfg.LogPrefix, err)  // line 91
}
...
log.Fatalf("%s: --headless-ensure: proxy did not become healthy within 10s", cfg.LogPrefix)  // line 107
```

`Ensure` is invoked from the SessionStart hook path. Any of these three failure modes — `os.Executable` failing, the subprocess refusing to start, or the proxy not becoming healthy in 10s — currently aborts the host process during session startup. The correct pattern is to return an error and let the caller decide whether to fatal.

### Step 1 — Fix `authcheck.IsAuthenticated` (and `EnsureAuthenticated`)

Replace the hardcoded `"databricks"` string in both `IsAuthenticated` and `EnsureAuthenticated` with a call to `resolveDatabricksCLI()` (already used in `token.go`). The resolution logic is currently in `package main` (`token.go`), so `pkg/authcheck` can't import it directly. Move `resolveDatabricksCLI`, `fallbackCLIDirs`, and `isExecutableFile` into a shared package (e.g. `pkg/cli` or `pkg/dbxcli`) and have both `main` and `pkg/authcheck` import it. While there, accept an explicit `cmdName` parameter on the authcheck functions (mirroring `NewTokenProvider(profile, cmdName string)` and `DiscoverHost(profile, cmdName string)`) so callers can override for tests and edge cases.

### Step 2 — Replace `log.Fatalf` in `pkg/proxy/proxy.go`

Change `NewServer` from `func NewServer(config *Config) http.Handler` to `func NewServer(config *Config) (http.Handler, error)`. Return wrapped errors at lines 282 and 287 instead of `log.Fatalf`. Update every call site (search `NewServer(` across the repo and the sibling binaries) to handle the returned error — at the program entrypoint, `log.Fatalf` is acceptable; everywhere else, propagate.

### Step 3 — Replace `log.Fatalf` in `pkg/headless/headless.go`

Change `Ensure(cfg Config)` to `Ensure(cfg Config) error`. Replace the three `log.Fatalf` calls at lines 80, 91, and 107 with wrapped error returns (preserving the existing refcount-release-on-failure logic). Update the SessionStart hook registration site to receive and surface that error — log it and exit there if appropriate, but do not fatal inside the library.

### Step 4 — Add tests

- Unit test for `IsAuthenticated` that injects a non-default CLI path (via the new `cmdName` argument and/or the existing `execCommand` test seam) and verifies it uses the resolved path rather than a hardcoded `"databricks"`. Cover the GUI-launched-PATH scenario by stubbing `exec.LookPath` to fail and verifying the fallback dirs are consulted.
- Test for `proxy.NewServer` that passes an invalid `InferenceUpstream` / `OTELUpstream` and asserts a non-nil error is returned rather than the test process exiting.
- Test for `headless.Ensure` that forces the health check to fail (e.g. by pointing `Port` at a closed port and stubbing the binary path to `/bin/true`) and asserts an error is returned rather than `os.Exit`.

### Step 5 — PR

Branch `fix/fatal-in-library-code`. Use a `fix:` conventional commit prefix so release-please picks it up. PR description should reference this issue and call out the breaking signature changes on `NewServer` and `Ensure` (callers in this repo and any consumers — `databricks-codex`, `databricks-opencode` per the headless.go package comment — must be updated in the same PR).

## Acceptance criteria

- `authcheck.IsAuthenticated` and `authcheck.EnsureAuthenticated` use the same CLI resolution as `token.go` (shared helper, no hardcoded `"databricks"` string).
- No `log.Fatalf` calls in `pkg/proxy` or `pkg/headless` (or any exported library function in this repo).
- All callers of `NewServer` and `Ensure` handle the returned errors gracefully; only `main` is allowed to fatal.
- New tests cover the resolution-path and error-return behaviors and pass on CI.

IceRhymers:
Fixes #120.

- `pkg/proxy.NewServer` now returns `(http.Handler, error)` — callers handle the error; only `main` fatals
- `pkg/headless.Ensure` now returns `error` — SessionStart hook path logs and exits cleanly
- `pkg/authcheck.IsAuthenticated` / `EnsureAuthenticated` accept explicit `cmdName` and use `pkg/cli.ResolveDatabricksCLI` — fixes silent false-negative on GUI launches where bare `databricks` is not on PATH
- New `pkg/cli` package extracted from `token.go` with resolution logic
- Tests verify error return (no process exit) for invalid upstream URLs and bad headless config

### DIFF
--- a/hooks.go
+++ b/hooks.go
@@ -16,12 +16,14 @@ import (
 // If not, it starts a detached headless proxy and polls until ready (max 10s).
 // Called by the SessionStart hook via: databricks-claude --headless-ensure
 func headlessEnsure(port int) {
-	headless.Ensure(headless.Config{
+	if err := headless.Ensure(headless.Config{
 		Port:          port,
 		ManagedEnvVar: "DATABRICKS_CLAUDE_MANAGED",
 		LogPrefix:     "databricks-claude",
 		RefcountPath:  refcount.PathForPort(".databricks-claude-sessions", port),
-	})
+	}); err != nil {
+		log.Fatalf("databricks-claude: %v", err)
+	}
 }
 
 // headlessRelease calls POST /shutdown on the proxy to decrement the refcount.
--- a/main.go
+++ b/main.go
@@ -275,7 +275,7 @@ func main() {
 	}
 
 	// --- Ensure the user is authenticated before proceeding ---
-	if err := authcheck.EnsureAuthenticated(resolvedProfile); err != nil {
+	if err := authcheck.EnsureAuthenticated(resolvedProfile, ""); err != nil {
 		log.Fatalf("databricks-claude: auth failed: %v", err)
 	}
 
--- a/pkg/authcheck/authcheck.go
+++ b/pkg/authcheck/authcheck.go
@@ -7,6 +7,8 @@ import (
 	"os/exec"
 	"strings"
 	"time"
+
+	"github.com/IceRhymers/databricks-claude/pkg/cli"
 )
 
 // Overridable for testing.
@@ -15,10 +17,12 @@ var execCommandContext = exec.CommandContext
 
 // IsAuthenticated returns true if a valid token can be fetched for the given
 // Databricks profile without triggering an interactive login.
-func IsAuthenticated(profile string) bool {
+// cmdName is the Databricks CLI binary name or path; pass "" to use the default ("databricks").
+func IsAuthenticated(profile, cmdName string) bool {
+	resolved := cli.ResolveDatabricksCLI(cmdName)
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
-	out, err := execCommandContext(ctx, "databricks", "auth", "token", "--profile", profile).Output()
+	out, err := execCommandContext(ctx, resolved, "auth", "token", "--profile", profile).Output()
 	if err != nil {
 		return false
 	}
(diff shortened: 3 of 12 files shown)


IceRhymers:
#claw no unit tests for the new cli pkg, should be included.

IceRhymers:
Added 9 tests in `pkg/cli/cli_test.go` covering `IsExecutableFile` (executable, non-executable, missing, directory) and `ResolveDatabricksCLI` (empty default, absolute passthrough, `DATABRICKS_CLI` env override, non-executable override ignored, fallback dir scan). All pass locally.

🦀 *(reply via claw)*

PR #152, opened 2026-05-11
chore: release 1.0.0

### LINKED ISSUE #150: feat(desktop): databricksCliPath MDM key for endpoint CLI discovery
## Summary

The credential helper resolves the `databricks` CLI binary via `state.DatabricksCLIPath → $DATABRICKS_CLI → exec.LookPath → FallbackCLIDirs scan → "databricks"` (final fallback exec failure). Under launchd's minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin` — what GUI subprocesses like Claude Desktop's credential-helper exec inherit), `exec.LookPath` will not find anything, leaving only the `FallbackCLIDirs` scan (`pkg/cli/cli.go:14-21`) to save the day.

This *probably* works on most endpoints (the scan covers `/usr/local/bin`, `/opt/homebrew/bin`, `~/.local/bin`, etc.) — but it's an implicit, undocumented, brittle fallback. MDM-deployed CLIs at non-standard paths (e.g. `/opt/databricks/bin/`) are invisible unless admins per-endpoint provision state, which defeats the entire point of MDM.

**Design principle**: the helper has exactly two resolved inputs — profile and CLI path. Profile is already MDM-keyed (#146). CLI path should be too. One mobileconfig payload, both required inputs declared.

## Background

`runCredentialHelper`'s only two resolved inputs are profile (covered) and CLI path (this issue). A theoretical third — the path to the credential helper binary itself — is encoded by Claude Desktop's `inferenceCredentialHelper` plist key as an absolute path; never depends on PATH discovery.

The fix mirrors #146 (profile MDM key) and #148 (DEFAULT sentinel) exactly:

- Reader extension in `pkg/cli.ResolveDatabricksCLI` adds an MDM tier
- Writer extension in `desktop_config.go:buildMobileconfig`/`buildRegFile`/`buildDevModeJSON` emits the new key
- Generator: `--databricks-cli-path` already lands in state; thread it through to the MDM payload too — symmetric with how `--profile` writes to both

### Lane 1 — Resolution chain placement

- **Lane 1A**: `state → $env → exec.LookPath → MDM → FallbackCLIDirs`. MDM as "admin's pinned value when nothing local exists, but LookPath beats it."
- **Lane 1B (final)**: `state → $env → MDM → exec.LookPath → FallbackCLIDirs`. **Symmetric with the `databricksProfile` chain** (`state → MDM → "DEFAULT"`). Explicit admin intent never overridden by accidental discovery.

Same lesson as #148's `"DEFAULT"` sentinel — never let a fallback short-circuit explicit configuration. If admin pinned `/opt/databricks/bin/databricks` in MDM, that's their intent; allowing `LookPath` to find `/usr/local/bin/databricks` first would silently override fleet policy on dual-installed machines.

### Lane 2 — Validation strategy

- **Lane 2A**: validate at read, fall through silently. Hard to debug.
- **Lane 2B**: trust blindly, let exec fail loudly.
- **Lane 2C (final)**: validate, log warning via `helperDebugLog`, fall through. Endpoint debugging surfaces the misconfig (`"MDM databricksCliPath=%q not executable, falling through"`); recovery still works.

Mirrors how `state.DatabricksCLIPath` is handled today.

## Approach: `databricksCliPath` in our existing MDM domain

Add one key to the MDM payload we already ship:

```xml
<!-- darwin: com.icerhymers.databricks-claude PayloadContent dict -->
<key>databricksProfile</key>
<string>databricks-ai-inference</string>
<key>databricksCliPath</key>
<string>/opt/databricks/bin/databricks</string>
```

```
; windows: HKCU\SOFTWARE\IceRhymers\databricks-claude
"databricksProfile"="databricks-ai-inference"
"databricksCliPath"="C:\\Program Files\\Databricks\\databricks.exe"
```

New helper CLI-resolution chain:

```
1. cmdName == absolute path or path-qualified → return as-is
2. $DATABRICKS_CLI env override (launchd never sets, but kept for back-compat with CLI invocations)
3. MDM databricksCliPath (NEW) — validated executable check before returning
4. exec.LookPath(cmdName)
5. FallbackCLIDirs scan
6. cmdName (final fallback — caller's exec fails with original error)
```

Generator threading: `desktop generate-config --databricks-cli-path /opt/databricks/bin/databricks` now writes to both `state.DatabricksCLIPath` (for the generator's own proxy use) AND the MDM payload (for endpoints) — same pattern as `--profile`.

## Implementation surface

| Area | Location |
|------|----------|
| Add MDM tier to CLI resolution chain | `pkg/cli/cli.go:ResolveDatabricksCLI` |
| `SetMDMReader` injection (avoid cross-pkg dep per `pkg/AGENTS.md`) | `pkg/cli/cli.go` |
| `pkg/mdmprofile.ReadKey(domain, key)` API; `Read(domain)` becomes shim | `pkg/mdmprofile/{darwin,windows,other}.go` |
| Wire `cli.SetMDMReader(mdmprofile.ReadKey)` at startup | `main.go` |
| Mobileconfig: emit `databricksCliPath` key | `desktop_config.go:buildMobileconfig` |
| Windows registry: emit `databricksCliPath` value | `desktop_config.go:buildRegFile` |
| Dev-mode JSON: emit `databricksCliPath` field | `desktop_config.go:buildDevModeJSON` |
| Generator threads `--databricks-cli-path` into MDM payload | `desktop_config.go:runGenerateDesktopConfig` + builder signatures |
| Helper logs MDM CLI path resolution outcomes | `desktop_config.go:runCredentialHelper` via `helperDebugLog` |
| Tests for chain ordering (state > MDM > LookPath > scan) | `pkg/cli/cli_test.go` |
| Tests for `databricksCliPath` emission across all three artifacts | `desktop_config_test.go` |
| Tests for `ReadKey` and the `Read` shim | `pkg/mdmprofile/*_test.go` |

## Acceptance criteria

- [ ] `pkg/cli.ResolveDatabricksCLI` consults MDM after `$DATABRICKS_CLI` and before `exec.LookPath`
- [ ] `pkg/cli` exposes `SetMDMReader(func(string,string)(string,error))`; default reader returns `("", nil)`; `main.go` wires `mdmprofile.ReadKey` at startup
- [ ] `pkg/mdmprofile` exposes `ReadKey(domain, key string) (string, error)` on darwin/windows; existing `Read(domain)` becomes a shim calling `ReadKey(domain, "databricksProfile")`
- [ ] When MDM `databricksCliPath` is set to an executable absolute path, `ResolveDatabricksCLI("")` returns that path
- [ ] When MDM `databricksCliPath` is set but the file is not executable, `ResolveDatabricksCLI("")` falls through to `exec.LookPath` then `FallbackCLIDirs`
- [ ] When MDM `databricksCliPath` is empty/unset, `ResolveDatabricksCLI("")` behaves identically to today (no regression)
- [ ] `state.DatabricksCLIPath` (when populated) still wins over MDM — symmetric with `databricksProfile` chain
- [ ] `runGenerateDesktopConfig --databricks-cli-path /opt/databricks/bin/databricks` writes that value into the mobileconfig second-payload `databricksCliPath` key
- [ ] Same value lands in the `.reg` output: `"databricksCliPath"="C:\\..."`
- [ ] Same value lands in dev-mode `.json` as a top-level `databricksCliPath` field
- [ ] `runGenerateDesktopConfig` without `--databricks-cli-path` **omits** the `databricksCliPath` key from all three artifacts (no empty-string emission)
- [ ] `helperDebugLog` records the MDM CLI path resolution step (consulted / succeeded / fell through)
- [ ] `GOOS=darwin go build ./...` and `GOOS=windows go build ./...` both succeed
- [ ] All packages pass `go test ./...`

## Out of scope

- **Migration of existing endpoints.** No users yet — clean-slate ship.
- **Anthropic-domain MDM key passthrough.** Their hardcoded values (`inferenceCredentialHelperTtlSec`, `disable*Telemetry`, `isClaudeCodeForDesktopEnabled`, etc.) are fleet-flexibility feature work, not bug-fix territory.
- **Native Desktop OTLP integration.** `otlpEndpoint`/`otlpProtocol`/`otlpHeaders` deserve their own feature work.
- **Helper-side validation of `state.DatabricksCLIPath`.** Today's silent fall-through preserved.
- **Speculation on future MDM keys.** Helper has only two resolved inputs; both covered after this PR. Don't add keys we don't need.

## Risks

- **`pkg/cli` cross-pkg dep on `pkg/mdmprofile`** would violate `pkg/AGENTS.md` ("Do not add cross-dependencies between sibling packages"). Mitigated by `SetMDMReader` injection — `pkg/cli` stays independently importable; main wires the real reader at startup. Same pattern as the test-overridable `mdmReader` var landed in #149.
- **`pkg/mdmprofile.ReadKey` is a new API surface.** Adding it is non-breaking. Refactoring `Read(domain)` into a shim that calls `ReadKey(domain, "databricksProfile")` is a one-line change that preserves all existing semantics.
- **Empty-vs-unset semantics in MDM payload.** Per acceptance criterion, omit the key entirely when CLI path isn't set. Test asserts no empty-string emission.

## Conventional commits

Single `feat:` commit (release-please minor bump):

```
feat(desktop): databricksCliPath MDM key for endpoint CLI discovery

The credential helper resolves the databricks CLI via state → $DATABRICKS_CLI →
exec.LookPath → FallbackCLIDirs scan. Under launchd's minimal PATH (GUI
subprocesses inherit /usr/bin:/bin:/usr/sbin:/sbin only), MDM-deployed CLIs at
non-standard paths (e.g. /opt/databricks/bin/) are invisible unless the admin
per-endpoint provisions state — which defeats the purpose of MDM.

Adds a databricksCliPath key to our existing MDM domain
(com.icerhymers.databricks-claude on darwin,
HKCU\SOFTWARE\IceRhymers\databricks-claude on windows). Helper's CLI resolution
chain becomes: state → $env → MDM → LookPath → FallbackCLIDirs. Symmetric with
the databricksProfile chain shipped in #146.

generate-config --databricks-cli-path now writes to both state (for the
generator) and the MDM payload (for endpoints) — same pattern as --profile.

pkg/mdmprofile gains ReadKey(domain, key) so multiple keys per domain work.
Existing Read(domain) becomes a shim. pkg/cli stays dependency-free via
SetMDMReader injection, populated by main at startup.

Refs: #<issue>
```

## Severity

**MEDIUM** — same urgency level as #148. No users today, but ships before MDM rollouts happen so admins never hit the implicit-fallback-scan surprise. ~30 LOC + 6 tests.

### LINKED ISSUE #153: feat(daemon): add databricks-claude serve subcommand for long-lived MDM-deployed instances
## Summary

Add a `databricks-claude serve` subcommand that runs the proxy as a long-lived, MDM/LaunchAgent-deployable daemon. Distinguished from the existing `--headless` per-session mode by: no refcount, no `/shutdown` route, append-only logging, and a `daemon: true` flag in `/health` so hooks (#154) can detect and no-op.

This is the foundation for the Claude Desktop Cowork-on-3P localhost-trusted deployment model, where the daemon owns Databricks OAuth refresh and Desktop sees only a fleet-shared fake key.

## Background — why a daemon

Claude Desktop's Cowork-on-3P managed configuration only accepts static `otlpHeaders`. Databricks OAuth refresh tokens expire roughly daily, so static-header auth from Desktop directly is a non-starter. The existing `--headless` per-session proxy solves this for CLI harnesses but isn't appropriate for Desktop (no per-session lifecycle, no hook surface).

A long-lived daemon owns the OAuth refresh loop, exposes inference + OTLP on `127.0.0.1`, and lets MDM ship a stable `gatewayBaseUrl` + fake `gatewayApiKey` to every endpoint without per-user secret distribution.

## Headless-auth validation (closes Q4 from design discussion)

Validated empirically before this issue was filed:

- macOS `nohup databricks auth login --profile <p> </dev/null` → browser pops, OAuth completes, no TTY needed.
- macOS LaunchAgent (`gui/<uid>` domain) invoking `databricks auth token` with `StandardInPath` unset → silent refresh of expired access token, fresh JWT in stdout, no browser pop, `expires_on` advances ~1h.

Conclusion: `EnsureAuthenticated` works as-is from a daemon. No new wakeup machinery needed.

## Decision (ADR)

- **Decision:** Add `databricks-claude serve` as a top-level subcommand alongside `desktop` and `setup`.
- **Drivers:** Discoverability via `--help`, parity with existing two-level subcommand pattern, clean plist invocation, room for future serve-specific flags.
- **Alternatives considered:** `--headless --daemon` flag-modifier on existing headless mode. Rejected: worse docs/discoverability story, awkward in LaunchAgent plist, doesn't match wrapper subcommand convention.
- **Consequences:**
  - Hooks need a `daemon: true` probe (handled in #154, not here).
  - `--headless` stays unchanged for backward compatibility.
  - `pkg/proxy` grows a `Daemon bool` config field to gate `/shutdown` registration and `/health` body shape.

## Implementation

1. **Add `serve` to the early-exit dispatcher in `main.go`** (hoisted above settings/state loads, per the wrapper's multi-entry-point pitfall — global wiring like `cli.SetMDMReader` must run before any dispatcher).
2. **New file `serve.go`** with `runServe(args []string) error`:
   - Flags: `--port`, `--log-file`, `--verbose`, `--profile`
   - Port: existing `pkg/state.ResolvePort` chain with `defaultPort=49153`
   - Profile: existing `flag → state → MDM → "DEFAULT"` resolution
   - Startup: `authcheck.EnsureAuthenticated(profile, "databricks-claude")` (existing function handles headless reauth as validated above)
   - Logging: `log.SetOutput(io.Discard)` default; `--verbose` adds stderr; `--log-file` opens with `os.O_APPEND|os.O_CREATE|os.O_WRONLY` (**not `O_TRUNC` — long-running daemon must preserve history**)
   - `os.Stdout = os.Stderr` at top of `runServe` to belt-and-suspenders prevent stdout pollution from SDK init
3. **`pkg/proxy` changes:**
   - Add `Daemon bool` to the proxy config
   - When `Daemon=true`: `/health` body returns `{"status":"ok","daemon":true,"version":"…","profile":"…","token_valid_until":"…"}`; `/shutdown` route is **not registered**
   - When `Daemon=false`: existing behavior unchanged
4. **`main.go:handleHelp()`** — add `serve` row to the subcommand list.
5. **`serve.go:printServeHelp()`** — usage/flags/examples/exit-codes mirroring `printDesktopHelp`.
6. **`README.md`** — new top-level `## serve Subcommand` section (not buried in an MDM use-case section, per the wrapper's four-docs-surfaces rule).
7. **`pkg/completion`** — register `serve` and its flags.
8. **Tests:**
   - `serve_test.go`: flag parsing, port resolution, log file opens in `O_APPEND` mode
   - `pkg/proxy/proxy_test.go`: `Daemon=true` → no `/shutdown` route (POST returns 404); `/health` body contains `daemon: true`
   - `pkg/proxy/proxy_test.go`: `Daemon=false` → existing test suite passes unchanged (regression guard)

## Acceptance criteria

- [ ] `databricks-claude serve --port 49153 --log-file /tmp/d.log` runs indefinitely
- [ ] Nothing written to stdout (verified by capturing for 10s)
- [ ] Log file written with `O_APPEND`; restart preserves prior content
- [ ] `curl http://127.0.0.1:49153/health` returns `200 {"daemon":true, …}`
- [ ] `curl -X POST http://127.0.0.1:49153/shutdown` returns `404`
- [ ] All existing `--headless` tests pass unchanged
- [ ] `databricks-claude --help` lists `serve`
- [ ] `databricks-claude serve --help` prints structured help
- [ ] README has top-level `## serve Subcommand` section
- [ ] Completion offers `serve` and its flags
- [ ] LaunchAgent smoke test: plist invoking `databricks-claude serve` runs, `/health` returns `daemon=true`

## Pre-mortem

1. **Stdout pollution from a transitive SDK call corrupts LaunchAgent stdout log.** Mitigation: `os.Stdout = os.Stderr` at the top of `runServe`. Test asserts stdout is empty after a brief run.
2. **Port collision after MDM rollout.** `portbind.Bind` falls back, but MDM-baked `gatewayBaseUrl` won't follow the fallback. Mitigation: `log.Fatal` on bind failure with the colliding-process PID; README documents that fleet port (49153) is owned by daemon.
3. **Refresh-token death mid-flight (>24h uptime).** Mitigation: call `EnsureAuthenticated` lazily on token-fetch failure in the proxy path, not just at startup. Tier-2 LaunchAgent test confirms headless recovery works.

## OTEL table routing in daemon mode

The wrapper's existing `--otel` / `--no-otel*` settings.json-mutating flags are **irrelevant to `serve`** — Claude Desktop reads OTLP config from MDM, not from any wrapper-mutated file. If the admin wants OTLP off fleet-wide, they omit `otlpEndpoint` from the MDM profile and the daemon never receives requests.

The **table-routing flags** (`--otel-metrics-table` / `--otel-logs-table` / `--otel-traces-table`) **are required** for `serve`. Without them, the proxy can't set `X-Databricks-UC-Table-Name` on outgoing OTLP requests (see `proxy.go:335`), and Databricks ingest 4xxs every OTLP request the daemon forwards. Shipping the daemon without these flags would ship a technically-running-but-functionally-broken v1.

### Flag surface on `serve`

- `--otel-metrics-table <fq-table>`
- `--otel-logs-table <fq-table>`
- `--otel-traces-table <fq-table>`

Flag names mirror the existing CLI flag names exactly, per the cross-wrapper symmetry rule.

### Resolution chain

```
flag → state (~/.claude/.databricks-claude.json) → MDM → empty
```

Same shape as `--profile` in the credential-helper path (per `pkg/mdmprofile/` and the precedent from #145/#149).

State file is **shared with CLI mode** — `state.OTELMetricsTable` / `state.OTELLogsTable` / `state.OTELTracesTable` are the same fields the CLI flow already persists. Rationale: matches how `Profile` is shared between CLI and credential-helper; MDM tier overrides state anyway for daemon-on-endpoint; on an admin's own laptop the daemon picking up their CLI prefs is usually what they want.

**State-tier sentinel guard required.** Apply the same `state != "" && state != "<sentinel>"` guard at every writer/reader of these fields, per the canonical pattern at `ensureconfig.go:42`. Without the guard, a stale `""` or accidentally-written sentinel beats the MDM tier and silently breaks endpoint deployments. Audit all three fields with `grep -n "state\.OTEL.*Table\s*=" --include="*.go"` and confirm every writer carries the guard.

### New MDM keys in own domain

Domain: `com.icerhymers.databricks-claude` (darwin plist) / `HKCU\SOFTWARE\IceRhymers\databricks-claude` (windows registry).

- `otelMetricsTable`
- `otelLogsTable`
- `otelTracesTable`

`pkg/mdmprofile.ReadKey` already exists from #149 and is wired via `cli.SetMDMReader` in the early-exit dispatcher. Just call it for these three additional keys from `serve`.

### Behavior when a signal's table is empty/unset

The daemon does **not** fail startup. It forwards OTLP for that signal without the `X-Databricks-UC-Table-Name` header. Databricks ingest will reject those requests, but that's the correct failure mode (visible, actionable, not silent). Document this in README.

### What NOT to add to `serve`

- `--otel` / `--no-otel` / `--no-otel-{metrics,logs,traces}` — these flags exist to mutate `~/.claude/settings.json`'s env block so Claude Code emits OTLP. Daemon is a forwarder; if MDM doesn't point Desktop at the OTLP listener, no traffic flows. These flags don't apply.

### Tests to add

- `serve_test.go`: resolution chain matrix for each of the three table fields — flag set / state set / MDM set / nothing set — confirming the right value lands in the proxy config.
- `serve_test.go`: state-tier sentinel guard regression — a state value of `""` falls through to MDM correctly.
- `pkg/proxy/proxy_test.go`: `Daemon=true` with empty table fields → outgoing OTLP request has no `X-Databricks-UC-Table-Name` header (regression guard against future "default to something" footguns).

## Out of scope

- Hook daemon-awareness (#154)
- OS service install/uninstall (#155)
- MDM profile generation for daemon mode (M-track, not yet filed)

## Follow-ups

- Log rotation (defer — `O_APPEND` is acceptable for v1; track separately if log volumes warrant it)

### LINKED ISSUE #154: feat(hooks): daemon-aware no-op behavior when serve is running
## Summary

When `databricks-claude serve` is running (introduced in the daemon issue), the existing `--headless-ensure` and `--headless-release` hook entry points must become side-effect-free no-ops: no refcount mutation, no `/shutdown` POST, no detached subprocess spawn. When no daemon is running, behavior is byte-identical to today.

**Depends on:** #153

## Decision (ADR)

- **Decision:** Add `pkg/health.ProxyMode(port, scheme) → "daemon" | "ephemeral" | ""` that consults `/health`'s JSON body. Hook entry points consult it first and short-circuit when mode == `"daemon"`.
- **Drivers:** Single source of truth (the running process), sub-100ms probe latency, room for future modes.
- **Alternatives considered:**
  - Separate `/daemon-mode` endpoint — rejected (two endpoints to keep in sync, extra round-trip, less curl-friendly for debugging).
  - Filesystem sentinel `/tmp/databricks-claude-daemon.lock` — rejected (race conditions on stale files after crash, cross-process correctness worse than HTTP).
- **Consequences:**
  - `pkg/health` grows a JSON-parsing function (currently 200-only).
  - `pkg/headless.Ensure` short-circuits when daemon detected — must skip refcount acquire.
  - `hooks.go:headlessRelease` short-circuits when daemon detected — must skip `/shutdown` POST.

## Implementation

1. **`pkg/health` — add `ProxyMode(port int, scheme string) (mode string, healthy bool)`:**
   - GET `<scheme>://127.0.0.1:<port>/health` with 500ms timeout
   - Returns `"daemon"` if response is 200 and JSON body contains `daemon: true`
   - Returns `"ephemeral"` if 200 with `daemon: false` or missing
   - Returns `""` if unreachable or non-200
2. **`pkg/headless/headless.go` daemon-aware short-circuit:**
   - Before `refcount.Acquire`, call `health.ProxyMode(cfg.Port, cfg.Scheme)`
   - If `"daemon"`: log "managed by daemon, hook is no-op" and return nil without acquiring refcount
   - Else: existing path unchanged
3. **`hooks.go:headlessRelease` daemon-aware short-circuit:**
   - Before POST `/shutdown`, call `health.ProxyMode(port, "http")`
   - If `"daemon"`: log no-op and return
   - Else: existing POST `/shutdown` unchanged
4. **Tests:**
   - `pkg/health/health_test.go`: `ProxyMode` covers daemon=true, daemon=false, malformed JSON, 200-empty, 500, unreachable
   - `hooks_test.go`: daemon-detected → refcount file unchanged, no shutdown POST observed
   - `hooks_test.go`: daemon-absent → all existing behavior preserved (regression guard)

## Acceptance criteria

- [ ] With `databricks-claude serve` running, `databricks-claude --headless-ensure` returns immediately, refcount file unchanged
- [ ] With `databricks-claude serve` running, `databricks-claude --headless-release` returns immediately, no `/shutdown` POST occurs (verified via daemon log)
- [ ] Without daemon, both entry points behave exactly as today (existing test suite passes unchanged)
- [ ] Hook completes in <200ms when daemon is up
- [ ] No new docs surfaces required (hook behavior is internal); optional one-liner in README troubleshooting

## Pre-mortem

1. **Slow `/health` under load → hook timeout.** Mitigation: probe inherits the existing 500ms timeout from `ProxyHealthy`; hook timeout budget is 15s (SessionStart) / 5s (SessionEnd).
2. **Daemon crashed but port still bound by zombie → ProxyMode returns "".** Hook falls through to today's `headless.Ensure`, which tries to spawn a child that fails to bind via `portbind`. Existing collision error surfaces.
3. **Mixed-version fleet: old wrapper hook + new daemon.** Old hook doesn't know about `daemon: true`, calls `/shutdown` → 404 (daemon doesn't register the route). Refcount never decremented by old hook (just logs). Daemon doesn't read refcount. Net result: safe degradation. Note in PR description.

## Out of scope

- Adding the daemon itself (#153)
- OS service install/uninstall (#155)

### LINKED ISSUE #155: feat(daemon): cross-platform service install/uninstall/status via serve subcommand
## Summary

Add `databricks-claude serve install`, `serve uninstall`, and `serve status` sub-subcommands that register the daemon as a per-user OS service on macOS, Windows, and Linux. Uses native primitives only: `launchctl` (macOS LaunchAgent), `schtasks.exe` (Windows Scheduled Task), `systemctl --user` (Linux). No brew/scoop dependency — the CLI owns the install story so any distribution mechanism works.

Canonical service name across all OSes: `databricks-claude-daemon`.

**Depends on:** #153

## Decision (ADR)

- **Decision:** `serve install` / `serve uninstall` / `serve status` as sub-subcommands under the `serve` noun.
- **Drivers:** One noun for run + management, parity with `desktop generate-config` two-level pattern, no new top-level surface, OS-native primitives only.
- **Alternatives considered:**
  - Top-level `daemon` noun separate from `serve` — rejected (two nouns for one concept).
  - Flags on `setup` (`--install-daemon` etc.) — rejected (three flags ≈ three sub-subcommands with worse ergonomics; `--daemon-status` as exit-code-only is awkward).
  - Homebrew `service do` block + Scoop `post_install` — rejected: CLI must own install so non-brew/non-scoop distributions work identically.
- **Per-user, not system-wide.** Daemon needs access to the user's OAuth keychain → must run in the user's session. No `sudo` / system-level `LaunchDaemon` / `systemctl` (system) paths.

## Per-platform matrix

| Concern | macOS | Windows | Linux |
|---|---|---|---|
| Mechanism | LaunchAgent | Scheduled Task (logon) | `systemd --user` |
| Manifest path | `~/Library/LaunchAgents/databricks-claude-daemon.plist` | (managed via `schtasks.exe`) | `~/.config/systemd/user/databricks-claude-daemon.service` |
| Install | write plist + `launchctl bootstrap gui/$(id -u) <plist>` + `kickstart -k …` | `schtasks /create /TN databricks-claude-daemon /SC ONLOGON /F /RL LIMITED /TR "<exe> serve …"` + `schtasks /run` | write unit + `systemctl --user daemon-reload` + `enable --now …` |
| Uninstall | `launchctl bootout gui/$(id -u)/databricks-claude-daemon` + `rm <plist>` | `schtasks /delete /TN databricks-claude-daemon /F` | `disable --now` + `rm <unit>` + `daemon-reload` |
| Status registered | `launchctl print gui/$(id -u)/databricks-claude-daemon` exit 0 | `schtasks /query /TN databricks-claude-daemon` exit 0 | `systemctl --user is-enabled` |
| Status running | `launchctl print` parsed for `state = running` | `schtasks /query /V` State column | `systemctl --user is-active` |
| Status healthy | All three: `pkg/health.ProxyMode(port)` returns `"daemon"` | | |
| Log path | `~/Library/Logs/databricks-claude-daemon/serve.log` | `%LOCALAPPDATA%\databricks-claude-daemon\serve.log` | `~/.local/state/databricks-claude-daemon/serve.log` |
| Binary path | All three: `os.Executable()` at install time, written into manifest | | |

## Implementation

1. **`serve_install.go` cross-platform dispatcher**
   - Parse sub-subcommand: `install` | `uninstall` | `status`
   - Flags: `--port` (default 49153), `--profile` (resolved via standard chain), `--log-file` (default per-OS)
   - Resolve binary via `os.Executable()` → absolute path baked into manifest
   - Dispatch via build tags to OS-specific impl
   - Pretty-print `status`: `Service: databricks-claude-daemon | Registered: yes | Running: yes | Healthy: yes (daemon=true, profile=<p>, expires_in=58m)`
2. **`serve_install_darwin.go`** — plist template, `launchctl bootstrap`/`kickstart`/`bootout`, parse `launchctl print` output. Plist includes `KeepAlive` dict, `RunAtLoad=true`, `WorkingDirectory`, stderr → logfile, stdout → /dev/null.
3. **`serve_install_windows.go`** — `schtasks /create … /SC ONLOGON /RL LIMITED`. Handle path escaping for spaces. `schtasks /query /V /FO CSV` parsing for status.
4. **`serve_install_linux.go`** — systemd unit with `Restart=on-failure`, `RestartSec=5`, `WantedBy=default.target`. `systemctl --user` orchestration.
5. **Tests:**
   - Unit: template rendering compared against golden files (plist XML, schtasks argv, systemd unit text)
   - Integration (CI per-OS): `install → status → uninstall → status` round-trip
   - macOS-specific: `plutil -lint` validates plist
6. **Docs (four surfaces, per skill):**
   - `handleHelp()` lists `serve install/uninstall/status`
   - `printServeHelp()` section per sub-subcommand
   - README sub-section under `## serve Subcommand`: per-OS one-liner table
   - `pkg/completion` registers `install`, `uninstall`, `status`

## Acceptance criteria

- [ ] `databricks-claude serve install` succeeds on macOS, Windows, Linux; daemon running within 5s
- [ ] `databricks-claude serve status` after install reports `Registered=yes, Running=yes, Healthy=yes`
- [ ] `databricks-claude serve uninstall` stops daemon and removes manifest; subsequent `status` reports `Registered=no, Running=no, Healthy=no`
- [ ] Re-running `install` after binary path changes rewrites manifest with new path (idempotent + path-refreshing)
- [ ] CI matrix passes on all three OSes
- [ ] README has per-OS one-liner table
- [ ] `serve --help` documents the three sub-subcommands

## Pre-mortem

1. **Stale manifest after binary upgrade.** Manifest references old binary path; daemon fails at next logon. Mitigation: `install` is idempotent and path-rewriting. `status` warns when manifest path ≠ current `os.Executable()`. README documents that upgrades should run `serve install` again.
2. **macOS Gatekeeper rejects unsigned local builds.** `launchctl bootstrap` may silently kill non-notarized binaries. Mitigation: `status` surfaces launchd's last exit code; `install` runs `spctl --assess` and warns when unsigned. Release-build notarization is a separate (existing) concern.
3. **Linux user-session expiry without lingering.** `systemd --user` services die at logout. For Claude Desktop usage this is irrelevant (user must be logged in to use it). Don't auto-call `loginctl enable-linger` (needs root, footgun). Document in README.

## Open question (resolve during implementation, not a blocker)

Should `serve install --profile <name>` bake profile into the manifest argv, or should the daemon read profile from MDM/state every startup? **Recommendation:** bake into manifest. MDM rollout emits manifest with right profile; solo-dev gets explicit control; matches "profile is sticky" UX.

## Out of scope

- The daemon process itself (#153)
- Hook daemon-awareness (#154)
- MDM profile generation for daemon mode (M-track, not yet filed)
- Log rotation (separate follow-up)

### LINKED ISSUE #162: feat(cli): --write-claude-config flag — scriptable settings.json bootstrap without starting the proxy
## Summary

Add a top-level `--write-claude-config` flag that runs the wrapper's first-run `settings.json` bootstrap and exits — no proxy startup, no port binding, no child process.

```bash
databricks-claude --write-claude-config
databricks-claude --write-claude-config --profile databricks-ai-inference
databricks-claude --write-claude-config --port 49153
```

## Motivation

This is the reconsidered Path B from #157. The originally-proposed `serve setup-claude-code` sub-subcommand turned out to be unnecessary: the wrapper's existing bootstrap path (`ensureConfig` → `needsFullSetup`) already writes permanent keys (it does NOT participate in the per-session Restore lifecycle), so a config-only invocation has zero new invariants to design around. Just expose what already exists.

A top-level flag (not a subcommand) is the right shape because:
- **Scriptable in MDM / fleet init.** `databricks-claude --write-claude-config --profile $PROF` is one line. No subcommand-nesting friction, no `setup` sub-subcommand to remember.
- **Parallel to existing flag-driven exits.** `--version`, `--print-env`, `--install-hooks`, `--headless` are all top-level flags that change wrapper behavior without a subcommand. `--write-claude-config` fits the same shape.
- **No port collision.** Today's workaround (`databricks-claude --headless` then Ctrl+C) requires binding the proxy port even though the user only wants the config write — a problem when a daemon already owns the port.

## Behaviour

1. Parse flags. If `--write-claude-config` is set, skip child-launch and proxy startup entirely.
2. Resolve profile (flag → saved state → MDM `databricksProfile` key → `"DEFAULT"`).
3. Resolve port (flag → saved state → default 49153). Do NOT bind it; just compute the URL.
4. Discover host via `databricks auth env` → build gateway URL (`{host}/ai-gateway/anthropic` for the upstream is only used internally; `ANTHROPIC_BASE_URL` written to settings is still `http://127.0.0.1:<port>`).
5. Call `ensureConfig(proxyURL, otelEnv)` with `needsFullSetup=true` so the full key set is written:
   - `ANTHROPIC_BASE_URL`
   - `ANTHROPIC_AUTH_TOKEN=proxy-managed`
   - `ANTHROPIC_MODEL` + the three `ANTHROPIC_DEFAULT_*_MODEL` keys
   - `ANTHROPIC_CUSTOM_HEADERS=x-databricks-use-coding-agent-mode: true`
   - `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`
6. Print a confirmation line to stderr (`databricks-claude: wrote env block to ~/.claude/settings.json (profile=<P>, base_url=<U>)`) and exit 0.

Idempotent — `ensureConfig` already short-circuits when the env block matches.

## Acceptance criteria

- [ ] `databricks-claude --write-claude-config` writes the full bootstrap env block to `~/.claude/settings.json` and exits 0 without binding any port or launching `claude`.
- [ ] `--profile` / `--port` are honoured when paired with `--write-claude-config`.
- [ ] Auth-failure path: if `databricks auth env` fails for the resolved profile, exit non-zero with the same actionable `Run 'databricks auth login --profile <name>' first` message used elsewhere — don't write a half-baked env block.
- [ ] OTEL flags (`--otel-metrics-table`, `--otel-logs-table`, `--otel-traces-table`) are honoured and persist via the existing state mechanism.
- [ ] `handleHelp` documents the flag and its purpose.
- [ ] Completion offers `--write-claude-config`.
- [ ] README "Using serve with Claude Code" section updated to recommend `--write-claude-config` over the `--headless` + Ctrl+C workaround.
- [ ] Test coverage: parser test (flag terminates early like `--print-env`/`--version`), an integration-style test that asserts the env block matches the `needsFullSetup` set after running the new code path.

## Out of scope

- An `--uninstall` or `--clear` counterpart. The wrapper doesn't currently own "remove these keys" as a first-class operation; if a user wants to back out, deleting the env block by hand or re-running `--install-hooks` flow is fine.
- Auto-detecting a running daemon's port. Port resolution stays driven by flag / saved state / default.

## References

- #157 (parent — Path A docs shipped in #161; this issue is the reconsidered Path B as a flag, not a subcommand)
- `main.go:738-746` for the `needsFullSetup` env-key set
- `ensureconfig.go` for the write path

### LINKED ISSUE #164: feat(desktop): --daemon mode for generate-config (gatewayBaseUrl + omitted helper)
## Summary

Add `--daemon` to `databricks-claude desktop generate-config` so MDM artifacts can point Claude Desktop at a local `databricks-claude serve` daemon instead of the Databricks AI Gateway directly.

**Default behavior is unchanged.** `desktop generate-config` (no flag) continues to emit helper-mode artifacts: `inferenceCredentialHelper` + `inferenceCredentialHelperTtlSec` + `gatewayBaseUrl` pointing at the official AI Gateway URL. This is the canonical, low-friction path for fleets that don't want to deploy a daemon. Helper-mode is the default forever — daemon-mode is a strict opt-in.

`--daemon` flips the emission to daemon-mode: omits `inferenceCredentialHelper`, sets `gatewayBaseUrl: http://127.0.0.1:<port>`, sets a fleet-shared static `gatewayApiKey` (localhost gate, not a real cred), and emits a new MDM payload in `com.icerhymers.databricks-claude` carrying the daemon port for endpoint verification.

## Motivation

Today's `desktop generate-config` only emits helper-mode artifacts. With the daemon track now shipped (#156 / #158 / #159), there's a second viable deployment shape for fleets: long-lived per-user daemon that owns OAuth refresh, serving Desktop via localhost.

Two viable inference paths going forward:

1. **Helper-mode (default).** Desktop talks directly to Databricks AI Gateway with per-call short-lived tokens from the credential helper. No daemon required. Lower complexity. **No OTLP from Desktop is possible** in this mode — Anthropic ships no `otlpCredentialHelper` (and is unlikely to — see May '26 design analysis), and OTLP requires dynamic auth that only a daemon can provide.
2. **Daemon-mode (`--daemon`).** Desktop talks to localhost daemon. Daemon owns OAuth refresh AND optional OTLP forwarding. Requires `databricks-claude serve install` on each endpoint. Higher deployment complexity, but the only way to do OTLP from Desktop.

The two modes are exclusive per-endpoint — an admin chooses one when generating the MDM artifact and pushes accordingly.

### Helper-mode (default — today's behavior, preserved verbatim)

| Key | Value |
|---|---|
| `inferenceCredentialHelper` | `/usr/local/bin/databricks-claude-credential-helper` (darwin) / equivalent path (windows) |
| `inferenceCredentialHelperTtlSec` | `55` |
| `gatewayBaseUrl` | resolved official AI Gateway URL for the profile (e.g. `https://<workspace>/ai-gateway/anthropic`) |
| `gatewayApiKey` | (absent — helper provides per-call token) |

Plus the existing `com.icerhymers.databricks-claude.databricksProfile` MDM payload from #149/#151 for endpoint profile resolution.

### Daemon-mode (`--daemon` opt-in)

| Key | Value |
|---|---|
| `inferenceCredentialHelper` | (OMITTED) |
| `inferenceCredentialHelperTtlSec` | (OMITTED) |
| `gatewayBaseUrl` | `http://127.0.0.1:<port>` (port from `state.Port` or `--port` override; default 49153) |
| `gatewayApiKey` | fleet-shared static fake key (configurable via `--daemon-fake-key`; default constant emitted with a banner warning) |
| `otlpEndpoint` | `http://127.0.0.1:<port>` (only if `--otel` also passed) |
| `otlpProtocol` | `http/protobuf` |
| `otlpHeaders` | content-type only (no auth header — daemon injects per-request) |

Plus a NEW MDM payload in `com.icerhymers.databricks-claude`:

| Key | Value |
|---|---|
| `daemonPort` | `<port>` (integer; endpoint `serve status` cross-checks) |
| `daemonMode` | `true` (boolean sentinel for any future endpoint-side conditional) |

The existing `databricksProfile` payload stays in both modes (daemon also resolves profile from MDM tier via `pkg/mdmprofile`).

## Flags

- `--daemon` — opt into daemon-mode emission. Default: helper-mode.
- `--port <N>` — override daemon port for the emitted `gatewayBaseUrl`. Default: `state.Port` → 49153.
- `--daemon-fake-key <K>` — override the static fake gateway key. Default: emit a banner warning + use a hard-coded constant.

## Acceptance criteria

- [ ] `databricks-claude desktop generate-config` (no flag) produces byte-identical output to today (helper-mode preserved).
- [ ] `databricks-claude desktop generate-config --daemon` emits the daemon-mode key set: `gatewayBaseUrl: http://127.0.0.1:<port>`, static `gatewayApiKey`, NO `inferenceCredentialHelper`.
- [ ] `--daemon --port 50000` honors the override in `gatewayBaseUrl`.
- [ ] `--daemon --otel` also emits OTLP keys pointing at `127.0.0.1:<port>`.
- [ ] `--daemon` emits the new `daemonPort` / `daemonMode` keys in the `com.icerhymers.databricks-claude` MDM payload.
- [ ] All three artifact formats (`.mobileconfig`, `.reg`, dev-mode `.json`) reflect the chosen mode consistently — no mode skew across files. A test asserts cross-file mode consistency.
- [ ] `handleHelp` documents the `--daemon` / `--port` / `--daemon-fake-key` flags.
- [ ] README "MDM / fleet rollout" section gains a one-liner pointing at the rollout guide (filed separately).
- [ ] Tests: matrix covering {helper-mode, daemon-mode} × {.mobileconfig, .reg, .json} × {with/without --otel} — assert key presence/absence per mode, not just emitted bytes.
- [ ] Cross-OS builds clean (`GOOS={darwin,windows,linux} go build ./...`).

## ADR

**Decision:** Add `--daemon` as opt-in flag, default to helper-mode emission.

**Drivers:**
1. Helper-mode is strictly simpler to deploy (no per-endpoint `serve install`). Most fleets won't want OTLP from Desktop on day one.
2. Daemon-mode is the only path to OTLP from Desktop — admins who need telemetry must opt in.
3. Single flag-flip across many MDM keys keeps the admin's mental model simple: "do I want telemetry? yes → `--daemon`, no → omit it."

**Alternatives considered:**
- *Default to daemon-mode.* Rejected — would silently raise the floor for all admins (every fleet now needs daemon install) and break existing rollouts.
- *Two separate subcommands (`generate-config-helper` / `generate-config-daemon`).* Rejected — duplicates the flag surface and obscures that the modes share most options.
- *Per-MDM-key flags (`--no-credential-helper`, `--use-localhost-gateway`).* Rejected — admin must remember which combination produces a coherent mode. Single `--daemon` flag emits the whole consistent set.

**Consequences:**
- Two emission paths must be tested with cross-file mode consistency (the matrix in ACs).
- The `desktop generate-config` help text gets one new "if you want OTLP, use `--daemon`" sentence.
- Endpoint-side: daemon must be installed (not handled by this issue — admin runs `serve install` per endpoint, possibly via MDM init script).

**Follow-ups:**
- MDM rollout guide for both modes (filed separately).
- Optional per-endpoint keychain bootstrap on first daemon start (filed separately).
- Future: when Anthropic ships `otlpCredentialHelper`, helper-mode + OTLP becomes possible and we can sunset the OTLP-requires-daemon constraint.

## Out of scope

- Per-user keychain bootstrap on first daemon start (filed separately).
- Notarization / Gatekeeper signing (#54 — daemon-mode rollout to real fleets wants signing first, but this issue is about emission, not deployment).
- System-wide `serve install` for cross-user MDM rollout (today `serve install` is per-user; cross-user documented as out of scope on #159).
- Endpoint-side daemon-mode detection / fallback logic (separate concern; daemon-mode is determined entirely by what the MDM artifact says).

## References

- May '26 daemon-mode design session — captured in `databricks-coding-agent-wrappers` skill `references/daemon-mode-design.md`
- #156 (daemon serve), #158 (daemon-aware hooks), #159 (serve install/uninstall/status) — closed daemon track
- #163 (`--write-claude-config`) — parallel Claude Code setup flow
- `pkg/mdmprofile/` — endpoint-side reader for the `com.icerhymers.databricks-claude` domain

### LINKED ISSUE #170: CLI #1: in-repo command-tree registry; migrate root command
Part of #169.

## Summary

Introduce an in-repo command-tree registry (`internal/cmd`) as the single source of truth for the CLI, and migrate the **root** command onto it. Help text and shell completion become *derived* from the tree instead of hand-maintained. No user-visible behavior change — same flags, same help content, same completion output.

## Motivation

Today the flag source-of-truth (`flagDefs` in `completion_flags.go`) is a flat list that only covers the root command, and help text is a hand-maintained `printf` in `handleHelp()`. Adding a flag means editing `flagDefs`, `parseArgs`, and `handleHelp()` by hand and hoping they stay in sync — only `knownFlags`↔`flagDefs` is tested. This issue replaces that with a declarative tree so help + completion + parse-metadata all derive from one declaration.

## Behaviour

1. New package `internal/cmd` defines:
   - `Command{ Name, Short, Long, Flags []FlagDef, Persistent []FlagDef, Subcommands []Command, Run func(...) error }`
   - `FlagDef` extended from today's completion `FlagDef` with resolution-chain metadata: `StateKey`, `EnvVar`, `MDMKey`, `Default` (so the flag→env→state→MDM→default chain can later be derived from the same source — not required to *use* it yet, just carry it).
2. The root command is declared as a `Command` value. `--profile` and `--port` are declared as **persistent** flags (inherited by every subcommand once #2 lands).
3. `handleHelp()` becomes a thin renderer over the root `Command` — the giant `printf` is deleted.
4. `pkg/completion` is fed from the tree (root command's `Flags` + child `Name`s) instead of the standalone `flagDefs`/`knownSubcommands` slices. The `--daemon`/`--daemon-fake-key` hack entries are removed from root flags (they belong to `desktop`, which moves in #2).
5. `parseArgs` reads `knownFlags` derived from the tree, not the flat slice. Transparent passthrough behavior (unknown flags → `claude`) is unchanged.

## Acceptance criteria

- [ ] `internal/cmd` package exists with `Command` + extended `FlagDef`; no dependency on the root `main` package (clean boundary — must be lift-out-able to `pkg/` later).
- [ ] Root command declared as a `Command` value; `handleHelp()` renders from it.
- [ ] `databricks-claude --help` output is byte-for-byte equivalent to today's (or intentionally improved — call out any diff in the PR).
- [ ] `databricks-claude completion bash|zsh|fish` output is functionally equivalent to today's for the root command.
- [ ] Parity test extended: every flag the binary parses has a corresponding tree node, AND every tree flag is parsed. Fails loudly on drift.
- [ ] `--profile` / `--port` declared persistent.
- [ ] `--daemon` / `--daemon-fake-key` removed from root flag set (no longer offered as root completions).
- [ ] All existing tests pass; no behavior change.

## Out of scope

- Migrating `serve`/`setup`/`desktop` onto the tree — that's #2.
- `config` / `hooks` subcommands — #3 / #4.
- Extracting `internal/cmd` to `pkg/` — deferred follow-up track.

### LINKED ISSUE #171: CLI #2: migrate serve/setup/desktop subcommands onto the command tree
Part of #169. Depends on #170.

## Summary

Migrate the existing subcommands — `serve` (+ `serve install|uninstall|status`), `setup`, and `desktop` — off their hand-rolled `flag.FlagSet` + `printXxxHelp()` functions and onto the `internal/cmd` command tree from #170. No user-visible behavior change.

## Motivation

After #170, the root command derives help + completion from the tree, but `serve`/`setup`/`desktop` still hand-roll everything:

- Each `runXxxCommand` builds its own `flag.FlagSet`.
- Each has a hand-maintained `printXxxHelp()` (`printServeHelp`, `printSetupHelp`, `printDesktopHelp`, `printServeInstallRootHelp`, `printServeInstallHelp`, `printServeUninstallHelp`, `printServeStatusHelp` — 7 blobs).
- Their flags are in **no** registry, so completion can't see them and there's no parity test.

This issue brings them under the same source of truth, kills the 7 help blobs, and gives nested completion (`serve install --<TAB>`) for free.

## Behaviour

1. `serve`, `setup`, `desktop` declared as `Subcommands` of the root `Command`. `serve install|uninstall|status` declared as nested `Subcommands` of `serve`.
2. `--daemon` / `--daemon-fake-key` declared on the `desktop` command (where they belong) instead of the removed root entries.
3. Each subcommand's `printXxxHelp()` deleted; help renders from its `Command` node.
4. `pkg/completion` walks the tree for nested completion — `serve <TAB>` offers `install/uninstall/status`, `serve install --<TAB>` offers that command's flags. (Completion package changes to support depth > 1 belong here.)
5. The hand-rolled `flag.FlagSet` parsing in each runner is replaced by the tree's parse path. Resolution chains (flag → state → MDM → default) preserved exactly — same precedence, same sentinels.
6. Parity test from #170 extended to cover all subcommand flags.

## Acceptance criteria

- [ ] `serve`, `setup`, `desktop` (+ `serve install|uninstall|status`) declared in the tree; runners parse via the tree.
- [ ] All 7 `printXxxHelp()` functions deleted; `--help` for each command renders from the tree and is functionally equivalent to today's output.
- [ ] `--daemon` / `--daemon-fake-key` live on `desktop`.
- [ ] Nested completion works: `databricks-claude serve <TAB>` → subcommands; `databricks-claude serve install --<TAB>` → flags.
- [ ] Resolution-chain behavior for `--profile` / `--port` / OTEL tables unchanged — verify with the existing resolution tests; add tests if coverage gaps surface.
- [ ] Parity test covers every subcommand flag.
- [ ] All existing tests pass; no behavior change.

## Out of scope

- `config` / `hooks` subcommands — #3 / #4.
- `--headless` / `serve` consolidation — #5.

IceRhymers:
:robot: I have created a release *beep* *boop*
---

### ⚠ BREAKING CHANGES

* removes the --headless and --idle-timeout root flags. The session-scoped standalone proxy (refcounted, /shutdown, idle-timeout, settings.json restore-on-exit) is now `databricks-claude serve --session-mode`. --idle-timeout becomes a `serve` flag. The long-lived daemon now requires explicit `serve --daemon` (bare `serve` with no mode flag or sub-subcommand is a hard error — eliminates silent-degradation hazard at the hooks spawn site). serve install|uninstall|status unchanged. The hooks deployment mode ([#173](https://github.com/IceRhymers/databricks-claude/issues/173)) now spawns `serve --session-mode` instead of `--headless`; the SessionStart/SessionEnd lifecycle is verified intact.
* removes 4 root flags. Replacements:   --install-hooks     → hooks install   --uninstall-hooks   → hooks uninstall   --headless-ensure   → hooks session-start  (hook-invoked internal)   --headless-release  → hooks session-end    (hook-invoked internal) Old flags are removed, not aliased. The generated SessionStart/SessionEnd hook JSON now invokes the new command names. No back-compat for already-installed hooks (none deployed). The hooks deployment mode itself is unchanged — proxy still spins up on SessionStart and tears down on SessionEnd.
* removes 14 root flags. Replacements:   --otel / --otel-*-table          → config otel enable [...]   --no-otel / --no-otel-{m,l,t}    → config otel disable [...]   --with-websearch / --websearch-* → config websearch enable [...]   --write-claude-config            → config write   --print-env                      → config show Old flags are removed, not aliased — they now pass through to claude. Storage semantics (two-store model, sentinel guards, state preservation on disable, OTEL section removal) are unchanged.

### Features

* **cli:** --write-claude-config flag for scriptable settings.json bootstrap ([d3b9a3b](https://github.com/IceRhymers/databricks-claude/commit/d3b9a3b83d11ad35c0922e727b4c20585ebd00d8))
* **cli:** --write-claude-config flag for scriptable settings.json bootstrap ([8da1957](https://github.com/IceRhymers/databricks-claude/commit/8da195783efd9a02bf232fafec3e95d59bd03db9)), closes [#162](https://github.com/IceRhymers/databricks-claude/issues/162)
* **cli:** scope --help to wrapper; advertise -- passthrough ([7d63ba4](https://github.com/IceRhymers/databricks-claude/commit/7d63ba4704101a680f9924f8a983826affa9755e))
* consolidate --headless proxy mode into `serve --session-mode` ([89c952a](https://github.com/IceRhymers/databricks-claude/commit/89c952a30c03e7e984bcceac1a658843235af3b6))
* consolidate hooks lifecycle flags under a `hooks` subcommand ([5b4a496](https://github.com/IceRhymers/databricks-claude/commit/5b4a4961ad203dacdd459d09b11acee2340ccafb))
* consolidate persistent-config flags under a `config` subcommand ([92b56d0](https://github.com/IceRhymers/databricks-claude/commit/92b56d01fd0c717551a9a02389c174ee566955fe))
* **daemon:** add serve subcommand for long-lived MDM-deployed instances ([52931b2](https://github.com/IceRhymers/databricks-claude/commit/52931b21aca3aad5d9e5f3a76f3f2eaf5ab88db0))
* **daemon:** add serve subcommand for long-lived MDM-deployed instances ([b6431af](https://github.com/IceRhymers/databricks-claude/commit/b6431afb7888e76829750656412cab5838af7b0c)), closes [#153](https://github.com/IceRhymers/databricks-claude/issues/153)
* **daemon:** cross-platform serve install/uninstall/status ([f8978f2](https://github.com/IceRhymers/databricks-claude/commit/f8978f245008d3dc48b6361ed4574418ed05fe1e))
* **daemon:** cross-platform serve install/uninstall/status ([b2a1f37](https://github.com/IceRhymers/databricks-claude/commit/b2a1f37e11df7824fff23a20e03950e822513d61)), closes [#155](https://github.com/IceRhymers/databricks-claude/issues/155)
* **desktop:** --daemon mode for generate-config ([03aa065](https://github.com/IceRhymers/databricks-claude/commit/03aa065cdd69b083616d7207b2dc1fdda4f62c3e)), closes [#164](https://github.com/IceRhymers/databricks-claude/issues/164)
* **desktop:** --daemon mode for generate-config (gatewayBaseUrl + omitted helper) ([b147c40](https://github.com/IceRhymers/databricks-claude/commit/b147c40c1992130226228e83ccd8cd1a3dc6f6a1))
* **desktop:** databricksCliPath MDM key for endpoint CLI discovery ([eac6ba7](https://github.com/IceRhymers/databricks-claude/commit/eac6ba7e06a1fb402b1a12be0cb54a9dc46e0be8))
* **desktop:** databricksCliPath MDM key for endpoint CLI discovery ([005eebd](https://github.com/IceRhymers/databricks-claude/commit/005eebd6a6d02f183c263fcbb810afe7f51024d5)), closes [#150](https://github.com/IceRhymers/databricks-claude/issues/150)
* **hooks:** daemon-aware no-op behavior when serve is running ([d314974](https://github.com/IceRhymers/databricks-claude/commit/d314974e718b768a2b18157ea777bce77df775cb))
* **hooks:** daemon-aware no-op behavior when serve is running ([1800b8d](https://github.com/IceRhymers/databricks-claude/commit/1800b8ddd5c3bf8c6de9fa861c938d654b85e6a8)), closes [#154](https://github.com/IceRhymers/databricks-claude/issues/154)
* introduce internal/cmd command-tree registry; migrate root command ([15068bb](https://github.com/IceRhymers/databricks-claude/commit/15068bb0776b9fd83cb9653f1d3682fed4364599)), closes [#170](https://github.com/IceRhymers/databricks-claude/issues/170)
* migrate serve/setup/desktop subcommands onto the command tree ([5579cfd](https://github.com/IceRhymers/databricks-claude/commit/5579cfdad2058a115a5577da60fe0eaff7bea4bd)), closes [#171](https://github.com/IceRhymers/databricks-claude/issues/171)

### Bug Fixes

* **build:** make Makefile work from PowerShell/cmd on Windows ([e4297ed](https://github.com/IceRhymers/databricks-claude/commit/e4297ede8a2cccb208bf0c97ddce3dd0d70142e5))
* daemon-mode otlpEndpoint must carry /otel path prefix ([b79739b](https://github.com/IceRhymers/databricks-claude/commit/b79739bb983ecdd4051c51240611c28e161938e4))
* **daemon:** address code review nits on PR [#159](https://github.com/IceRhymers/databricks-claude/issues/159) follow-up ([23b02f3](https://github.com/IceRhymers/databricks-claude/commit/23b02f3755d66b00e68bc2b9ce1278e38f71fedb))
* **daemon:** install-time pre-auth + daemon-safe serve + status accuracy ([6cedf58](https://github.com/IceRhymers/databricks-claude/commit/6cedf5814b8d43b03c7a34c998cfc34f6d2c0a18))
* **serve:** wire --with-websearch into daemon proxy config ([4f9c2a2](https://github.com/IceRhymers/databricks-claude/commit/4f9c2a294264717d9288b7876d51a0ea2e051d46))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).


IceRhymers:
🤖 Created releases:

- [v1.0.0](https://github.com/IceRhymers/databricks-claude/releases/tag/v1.0.0)

:sunflower:

PR #156, opened 2026-05-12
feat(daemon): add serve subcommand for long-lived MDM-deployed instances

IceRhymers:
## Summary

Implements #153 — long-lived `serve` subcommand for Claude Desktop Cowork-on-3P localhost-trusted deployment. The daemon owns Databricks OAuth refresh, exposes inference + OTLP on `127.0.0.1`, and lets MDM ship a stable `gatewayBaseUrl` + fake `gatewayApiKey` to every endpoint without per-user secret distribution.

Distinguished from existing `--headless` per-session mode by:
- No refcount (daemon is not session-managed)
- Explicit `/shutdown` 404 handler (route is rejected, not forwarded upstream)
- Append-only logging (`O_APPEND`, never `O_TRUNC`)
- `daemon: true` flag in `/health` so hooks (#154) can detect and no-op

## Headless reauth pre-verified

Before implementation, validated empirically on macOS that `databricks auth token` works fully headless under LaunchAgent (`gui/<uid>` domain) — silent refresh of an expired access token via `StandardInPath` unset, no TTY needed. Documented in the skill. No new wakeup machinery required.

## Pipeline

Implemented via OMC: ralplan plan (issue #153) → autopilot (88 turns, 15 min, $3.88) → cross-lane review round 1 (REQUEST_CHANGES: 2 MAJOR + 1 MINOR) → manual amend (no autopilot rerun) → cross-lane review round 2 (APPROVE).

### Round-1 review findings (all resolved before merge)

- **MAJOR — `/shutdown` did not actually return 404 in daemon mode.** The original implementation relied on the route being unregistered, but the inference catch-all (`mux.Handle("/", …)`) was forwarding `POST /shutdown` upstream to the Databricks AI Gateway. Fix: explicit `mux.HandleFunc("/shutdown", http.NotFound)` gated on `config.Daemon=true`, registered BEFORE the catch-all. New test `TestProxy_NonDaemon_NoExplicit404` confirms the 404 doesn't leak into the lifecycle-wrapper path.
- **MAJOR — AC #2 (stdout silence) was unimplemented in tests.** The implementation had `os.Stdout = os.Stderr` belt-and-suspenders at the top of `runServe`, but no test asserted stdout stays empty. Fix: `TestServe_StdoutSilence_HelpPath` builds the real binary, runs `serve --help`, asserts `stdout.Len() == 0`.
- **MINOR — sentinel-guard test was tautological** (identical to the MDM-wins-when-no-flag test). Fix: extracted writer guard into `shouldPersistOTELTable(flagSet, resolved, stateVal)` helper and added `TestShouldPersistOTELTable` with 6 subcases exercising the writer side directly, including explicit regressions for the PR #149 shadow-MDM footgun.

## Acceptance criteria from #153

- [x] `databricks-claude serve --port 49153 --log-file /tmp/d.log` runs indefinitely
- [x] Nothing written to stdout (verified in `TestServe_StdoutSilence_HelpPath`)
- [x] Log file written with `O_APPEND`; restart preserves prior content (verified in `TestParseServeFlags_LogFile`)
- [x] `curl http://127.0.0.1:49153/health` returns `200 {"daemon":true, …}` (verified in `TestProxy_Daemon_HealthReturnsDaemonTrue`)
- [x] `curl -X POST http://127.0.0.1:49153/shutdown` returns `404` (verified in `TestProxy_Daemon_ShutdownReturns404`, asserting both status and no-upstream-leak)
- [x] All existing `--headless` tests pass unchanged (no test modifications outside the new daemon-mode additions)
- [x] `databricks-claude --help` lists `serve`
- [x] `databricks-claude serve --help` prints structured help
- [x] README has top-level `## serve Subcommand` section
- [x] Completion offers `serve` and its flags (via new `SubcommandDef` registry in `pkg/completion`)

## OTEL table routing

Three flags on `serve` — `--otel-metrics-table`, `--otel-logs-table`, `--otel-traces-table` — with resolution chain `flag → state → MDM → empty`. New MDM keys in `com.icerhymers.databricks-claude`: `otelMetricsTable`, `otelLogsTable`, `otelTracesTable`. State fields shared with CLI mode per the design discussion (matches how `Profile` is shared).

The settings.json-mutating flags (`--otel`, `--no-otel*`) are NOT added to `serve` — those exist to patch `~/.claude/settings.json` for Claude Code, irrelevant in daemon mode where Desktop reads OTLP config from MDM.

Empty table = no `X-Databricks-UC-Table-Name` header; daemon does NOT fail startup (regression-tested in `TestProxy_Daemon_EmptyOTELTables`).

## Architecture invariants preserved

- Zero external Go deps (stdlib only).
- `pkg/` cross-sibling-dependency rule respected (`serve.go` lives in `main`, not in `pkg/`).
- Global wiring (`cli.SetMDMReader`) hoisted above all early-exit dispatchers in `main.go` per the multi-entry-point pitfall.
- Sentinel-guard pattern applied at every OTEL-table state writer via shared `shouldPersistOTELTable` helper.
- Four docs surfaces updated: `handleHelp`, `printServeHelp`, top-level README section, `pkg/completion` registry.

## Follow-ups (non-blocking)

- Log rotation (the daemon uses `O_APPEND`; rotation can be deferred to a future issue if log volumes warrant).
- Cosmetic: `runServe` returns `void` instead of `error`; uses `log.Fatalf` for fatal paths (functionally equivalent to the spec's signature, slightly weaker testability).
- The stdout-silence test verifies the outcome (stdout empty) for the `--help` path but not the specific `os.Stdout = os.Stderr` mechanism — would benefit from a `fmt.Println("contamination")` probe in a future test if daemon-startup integration testing becomes feasible without Databricks auth.

Refs #153

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,10 +31,11 @@ The root package is a set of `.go` files that act as thin facades wiring togethe
 - **state.go** — `persistentState` struct: JSON schema for `~/.claude/.databricks-claude.json` (profile, port, CLI path, OTEL table names).
 - **hooks.go** — Session hook install/uninstall (`installHooks`, `uninstallHooks`).
 - **ensureconfig.go** — Bootstrap helpers for first-run settings setup.
-- **completion_flags.go** — `flagDefs` slice driving shell completion and flag parsing.
+- **completion_flags.go** — `flagDefs` slice driving shell completion and flag parsing; `knownSubcommands` slice driving subcommand-name completion.
 - **databrickscfg.go** — Reads `~/.databrickscfg` section headers for profile name resolution.
 - **desktop_config.go** — `desktop` subcommand: `generate-config` writing `.mobileconfig`, `.reg`, and `.json` artifacts for Claude Desktop.
 - **desktop_trust.go** — `generate-trust-profile` subcommand for MDM trust profile generation.
+- **serve.go** — `serve` subcommand: long-lived daemon that serves Claude Code and Claude Desktop with persistent Databricks OAuth. A third deployment mode alongside the per-session CLI wrapper (`databricks-claude claude …`) and SessionStart hooks — useful when you want a single OAuth-refreshing proxy that survives across sessions. `runServe(args)` parses flags, resolves profile/port/OTEL tables, calls `authcheck.EnsureAuthenticated`, binds port exclusively, and runs the proxy with `Daemon=true`. `printServeHelp()` mirrors `printDesktopHelp`/`printSetupHelp` in shape.
 
 ### Library packages (`pkg/`)
 
--- a/README.md
+++ b/README.md
@@ -397,6 +397,65 @@ databricks-claude setup --profile databricks-ai-inference --force
 
 `setup` is the same auth flow the credential helper uses for daily token recovery — running it proactively in a fleet init script keeps users from seeing the recovery browser tab on their first Claude Desktop launch.
 
+## serve Subcommand
+
+Long-lived daemon that serves **both Claude Code and Claude Desktop** with persistent Databricks OAuth. A third deployment mode alongside the per-session CLI wrapper (`databricks-claude claude …`) and SessionStart hooks — useful when you want a single OAuth-refreshing proxy that survives across sessions.
+
+Owns Databricks OAuth refresh and exposes inference + OTLP on `127.0.0.1`. Distinguished from `--headless` mode by: no session refcount, no `/shutdown` route, append-only logging, and `daemon:true` in `/health` so hooks can detect and no-op.
+
+Designed for LaunchAgent (macOS) or systemd (Linux) deployment, where the daemon is started once at login and kept running. Configure your client to point at the daemon:
+- **Claude Desktop:** via MDM, set `gatewayBaseUrl: http://127.0.0.1:<port>` with a static fake API key (no per-user secret distribution).
+- **Claude Code:** edit `~/.claude/settings.json` once to set `ANTHROPIC_BASE_URL=http://127.0.0.1:<port>` in the env block. The daemon does NOT mutate `settings.json` itself — it stays outside the per-tool lifecycle by design.
+
+```bash
+# Minimal daemon on default port:
+databricks-claude serve
+
+# With explicit profile, port, and persistent log file:
+databricks-claude serve \
+  --profile databricks-ai-inference \
+  --port 49153 \
+  --log-file /var/log/databricks-claude/daemon.log
+
+# With OTEL table routing:
+databricks-claude serve \
+  --otel-metrics-table main.claude_telemetry.claude_otel_metrics \
+  --otel-logs-table main.claude_telemetry.claude_otel_logs
+```
+
+| Flag | Purpose |
+|------|---------|
+| `--port int` | Proxy listen port (default: `49153`). Bound exclusively — MDM-baked `gatewayBaseUrl` is a fixed URL and cannot follow a fallback port. |
+| `--profile string` | Databricks config profile (default: saved state → MDM `databricksProfile` key → `"DEFAULT"`) |
+| `--log-file string` | Append-only log file (`O_APPEND`, not `O_TRUNC`). Safe for log rotation. Restarts preserve prior content. |
+| `--verbose`, `-v` | Also write debug logs to stderr (combinable with `--log-file`) |
+| `--otel-metrics-table string` | Unity Catalog table for OTEL metrics. Resolution: flag → saved state → MDM `otelMetricsTable` key → empty. |
+| `--otel-logs-table string` | Unity Catalog table for OTEL logs (same resolution chain) |
+| `--otel-traces-table string` | Unity Catalog table for OTEL traces (same resolution chain) |
+| `--help`, `-h` | Show subcommand help |
+
+**OTEL table behavior when empty:** The daemon does **not** fail startup when a table is unset. It forwards OTLP for that signal without the `X-Databricks-UC-Table-Name` header; Databricks ingest rejects those requests with a visible 4xx — an actionable failure, not a silent one.
+
+**MDM keys** (domain `com.icerhymers.databricks-claude`):
+
+| Key | Purpose |
+|-----|---------|
+| `databricksProfile` | Databricks CLI profile name |
+| `otelMetricsTable` | UC table for OTEL metrics |
+| `otelLogsTable` | UC table for OTEL logs |
+| `otelTracesTable` | UC table for OTEL traces |
+
+**Endpoints:**
+
+| Endpoint | Description |
+|----------|-------------|
+| `GET /health` | Returns `{"tool":"databricks-claude","daemon":true,"version":"...","profile":"...","token_valid_until":"..."}` |
+| `POST /shutdown` | **Not registered** — returns 404. Stop the daemon via SIGTERM (e.g. `launchctl stop` or `systemctl stop`). |
+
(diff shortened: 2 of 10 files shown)


CLAUDE.md
```diff
@@ -31,10 +31,11 @@ The root package is a set of `.go` files that act as thin facades wiring togethe
...
 - **hooks.go** — Session hook install/uninstall (`installHooks`, `uninstallHooks`).
 - **ensureconfig.go** — Bootstrap helpers for first-run settings setup.
-- **completion_flags.go** — `flagDefs` slice driving shell completion and flag parsing.
+- **completion_flags.go** — `flagDefs` slice driving shell completion and flag parsing; `knownSubcommands` slice driving subcommand-name completion.
 - **databrickscfg.go** — Reads `~/.databrickscfg` section headers for profile name resolution.
 - **desktop_config.go** — `desktop` subcommand: `generate-config` writing `.mobileconfig`, `.reg`, and `.json` artifacts for Claude Desktop.
 - **desktop_trust.go** — `generate-trust-profile` subcommand for MDM trust profile generation.
+- **serve.go** — `serve` subcommand: long-lived daemon for MDM-deployed Claude Desktop deployments. `runServe(args)` parses flags, resolves profile/port/OTEL tables, calls `authcheck.EnsureAuthenticated`, binds port exclusively, and runs the proxy with `Daemon=true`. `printServeHelp()` mirrors `printDesktopHelp`/`printSetupHelp` in shape.
```
IceRhymers: I wouldn't say that this is for just MDM Desktop deployments. This would work with claude code when configured at least once.

main.go
```diff
@@ -1083,6 +1092,10 @@ Subcommands:
                                runs 'databricks auth login' when not authenticated. Designed
                                for fleet init scripts and per-user login agents.
                                Run 'databricks-claude setup --help' for flags.
+  serve [flags]                Long-lived daemon for MDM-deployed Claude Desktop deployments.
```
IceRhymers: Same deal here, for both claude desktop and code

README.md
```diff
@@ -397,6 +397,61 @@ databricks-claude setup --profile databricks-ai-inference --force
 
 `setup` is the same auth flow the credential helper uses for daily token recovery — running it proactively in a fleet init script keeps users from seeing the recovery browser tab on their first Claude Desktop launch.
 
+## serve Subcommand
+
+Long-lived daemon for MDM-deployed Claude Desktop deployments. Owns Databricks OAuth refresh and exposes inference + OTLP on `127.0.0.1`. Distinguished from `--headless` mode by: no session refcount, no `/shutdown` route, append-only logging, and `daemon:true` in `/health` so hooks can detect and no-op.
```
IceRhymers: This one too, this serve is for both desktop and claude code.

serve.go
```diff
@@ -0,0 +1,363 @@
...
+	signal.Stop(sigCh)
+	log.Printf("databricks-claude: serve: received %s, shutting down", sig)
+	ln.Close()
+}
+
+// printServeHelp prints usage for the `serve` subcommand to stderr.
+func printServeHelp() {
+	fmt.Fprint(os.Stderr, `Usage: databricks-claude serve [flags]
+
+Long-lived daemon for MDM-deployed Claude Desktop deployments.
```
IceRhymers: Here too

serve.go
```diff
@@ -0,0 +1,363 @@
...
+MDM keys (domain: com.icerhymers.databricks-claude):
+  databricksProfile   Databricks CLI profile name
+  otelMetricsTable    UC table for OTEL metrics
+  otelLogsTable       UC table for OTEL logs
+  otelTracesTable     UC table for OTEL traces
+
+Note: --otel / --no-otel* flags are NOT supported for serve. Those flags
+mutate ~/.claude/settings.json to configure Claude Code's OTLP emission.
+In daemon mode, Claude Desktop reads OTLP config from MDM, not from any
+wrapper-mutated file. Omit otlpEndpoint from the MDM profile to disable OTLP.
```
IceRhymers, line 344: Maybe we need to add a guide for to how to enable telemetry properly, so you don't need to use the flags each time if you only want to use deamon mode. I think that the deamon shouldn't write the ~/.claude/settings.json because it's long living, and outside of the lifecycle of the tools that actually use it. But people will want to set up this once, and then just use `claude` or claude desktop. This deamon serves as anohter way to serve claude with databrciks and use oauth that isn't the direct CLI wrapper (databricks-claude) or the hooks we have.

CLAUDE.md
IceRhymers: Good catch — reframed in `b6431af`. `CLAUDE.md` now describes the daemon as serving both Claude Code and Claude Desktop, framed as a third deployment mode alongside the per-session CLI wrapper and SessionStart hooks.

main.go
IceRhymers: Reframed in `b6431af`. `handleHelp()`'s serve row now reads "Long-lived daemon serving Claude Code and Claude Desktop with persistent Databricks OAuth … A third deployment mode alongside the per-session CLI wrapper and SessionStart hooks."

README.md
IceRhymers: Reframed in `b6431af`. The README section opener now leads with "Long-lived daemon that serves **both Claude Code and Claude Desktop** with persistent Databricks OAuth. A third deployment mode alongside the per-session CLI wrapper (`databricks-claude claude …`) and SessionStart hooks." Also added a per-client setup blurb covering both Desktop (MDM) and Code (manual settings.json edit) with a callout that the daemon does NOT mutate settings.json by design.

serve.go
IceRhymers: Reframed in `b6431af`. `printServeHelp()` now opens with "Long-lived daemon that serves Claude Code and Claude Desktop with persistent Databricks OAuth. A third deployment mode alongside the per-session CLI wrapper (`databricks-claude claude ...`) and SessionStart hooks — useful when you want a single OAuth-refreshing proxy that survives across sessions." The "Designed for LaunchAgent or systemd" block was extended with per-client setup pointers (Desktop = MDM `gatewayBaseUrl`; Code = manual `ANTHROPIC_BASE_URL` in settings.json) plus the daemon-doesn't-mutate-settings.json invariant called out explicitly.

serve.go
IceRhymers: Agreed on both the principle and the scope. Filed as #157 — the daemon stays outside the settings.json lifecycle (won't write `~/.claude/settings.json`), and the one-time client setup gets its own surface. Scoped narrowly per your guidance: just Claude Code (Desktop is already MDM-driven), no overengineered enable-client noun. Path A (README-only) vs Path B (`serve setup-claude-code` subcommand) is decided in that issue, not this PR. Note that OTEL table flags already persist to state on first invocation via `shouldPersistOTELTable`, so the "set once, omit thereafter" UX is already there for those — the missing piece is just the client-side `ANTHROPIC_BASE_URL` config.

IceRhymers:
Addressed review comments in `b6431af`:

- **Reframe (comments 1-4):** Updated CLAUDE.md, `handleHelp()`, README, and `printServeHelp()` to frame `serve` as a third deployment mode serving **both Claude Code and Claude Desktop** with persistent Databricks OAuth, alongside the per-session CLI wrapper and SessionStart hooks. The README and `printServeHelp` also gained per-client setup pointers (Desktop = MDM `gatewayBaseUrl`; Code = manual settings.json edit) and an explicit note that the daemon does NOT mutate `~/.claude/settings.json` by design.
- **Telemetry/setup guide (comment 5):** Filed as #157. Kept narrow per discussion — just Claude Code (Desktop is already MDM-driven), no overengineered enable-client noun. The architectural question (permanent-vs-lifecycle settings.json patch) is captured there for follow-up. OTEL table flags already persist to state on first invocation via `shouldPersistOTELTable`, so the "set once, omit thereafter" UX is in place for those — only `ANTHROPIC_BASE_URL` client config is the missing piece.

Tests still green (`gofmt`, `go vet`, `go build`, full `go test ./...` suite). Single commit preserved via `--amend`.

PR #159, opened 2026-05-12
feat(daemon): cross-platform serve install/uninstall/status

IceRhymers:
## Summary

Implements #155 — three sub-subcommands under the existing `serve` noun for one-command OS-service registration: `install`, `uninstall`, `status`. Native primitives only (`launchctl` / `schtasks` / `systemctl --user`), per-user only, no sudo, no third-party service managers.

Canonical service name across all OSes: `databricks-claude-daemon`.

This completes the daemon track. With #156 (D1 daemon), #158 (D2 daemon-aware hooks), and now this, `databricks-claude` covers all four reasonable deployment shapes: per-session CLI wrapper, SessionStart hooks, MDM-driven Desktop, and daemon-as-service for solo/OSS users.

## Per-OS implementation

| OS | Mechanism | Manifest path |
|---|---|---|
| macOS | LaunchAgent | `~/Library/LaunchAgents/databricks-claude-daemon.plist` |
| Windows | Scheduled Task (logon) | (managed via `schtasks.exe`) |
| Linux | `systemd --user` | `~/.config/systemd/user/databricks-claude-daemon.service` |

**`install`** writes the manifest with the absolute binary path from `os.Executable()` baked in, then starts the service. Idempotent — re-running rewrites the manifest with the current binary path (covers binary-upgrade case).

**`uninstall`** stops AND removes the registration in one shot. Tolerates "not installed" gracefully on all three OSes.

**`status`** reports Registered / Running / Healthy in one command. Healthy probe reuses `pkg/health.ProxyMode` from #158.

## Per-user clarifications (from PR discussion)

- **Linux user-session expiry** — `systemd --user` services die at logout. No auto-call to `loginctl enable-linger` (needs root). Documented in README as the user's responsibility if they want the daemon to survive logout. For Claude Code / Desktop usage this is moot — users are logged in when using the client.
- **macOS unsigned binaries** — `install` runs `spctl --assess --type execute <binary>` and prints a one-line warning if it fails, but the install proceeds. Locally-built unsigned binaries may be silently killed by `launchctl bootstrap`, but technical users running self-built proxies should be aware.

## Pipeline

Implemented via OMC: autopilot (73 turns, 9.7 min, $3.02) → **three rounds of review**.

### Round 1: REQUEST_CHANGES (1 BLOCKER + 1 MAJOR)

- **BLOCKER:** `serve <sub> --help` reached the parent `printServeHelp` because `hasFlag("--help")` short-circuited BEFORE the sub-subcommand dispatch in `runServe`. The dedicated `printServe{Install,Uninstall,Status}Help` functions were dead code on the user-facing path.
- **MAJOR:** `serve status` output landed on stderr because `runServe` reassigned `os.Stdout = os.Stderr` before the sub-subcommand dispatch — making `serve status | grep Healthy` broken in shipped binaries.

Fix: hoisted the sub-subcommand dispatch ABOVE both the `--help` short-circuit AND the stdout redirect in `runServe`. Added two subprocess regression tests.

### Round 2: REQUEST_CHANGES (1 MAJOR — test-fidelity)

Reviewer correctly caught that the new `TestServe_SubSubcommandHelp_RoutesCorrectly` sentinels (`"serve install"`, etc.) ALSO appeared in the parent `printServeHelp` body's "Sub-subcommands" and "Examples" sections — so the test would still pass under the round-1 regression (false-passing guard).

Fix: tightened sentinels to strings genuinely unique to each sub-subcommand help (`"Service name: databricks-claude-daemon"`, `Tolerates "not installed" gracefully.`, `"Registered — manifest/task/unit file exists"`). Added a parent-body capture and `t.Fatalf` guard at the top of each subcase that detects future help-text drift bringing a sentinel back into the parent — guard-the-guard mechanism.

I empirically validated the fix by temporarily reverting the dispatch hoist in `serve.go` and confirming all four affected subtests failed cleanly, then restored.

### Round 3: APPROVE

Reviewer verified the three sentinels appear at exactly one location each (`serve_install.go:346/369/381`), do NOT appear in `printServeHelp` (`serve.go:309-409`), and the guard-the-guard mechanism works bidirectionally. Round-1 fixes still hold. All gates green across the three-OS build matrix.

## Acceptance criteria from #155

- [x] `databricks-claude serve install` succeeds on macOS / Windows / Linux (cross-compile validated — integration tests against real service managers explicitly out of scope for CI)
- [x] `serve status` after install would report `Registered=yes, Running=yes, Healthy=yes`
- [x] `serve uninstall` stops the daemon and removes the manifest
- [x] Re-running `install` after binary path changes rewrites the manifest (idempotent + path-refreshing)
- [x] CI gates pass on all three OSes (`go vet`, `gofmt`, `GOOS={linux,darwin,windows} go build`, full test suite)
- [x] README has per-OS one-liner table
- [x] `serve --help` documents the three sub-subcommands; `serve install --help` etc. reach the right help (regression-tested)

## Non-blocking nits deferred (from round-1)

- `--port` `strconv.Atoi` silently swallows errors (`f.port, _ = ...`)
- Linux `disable --now` tolerance uses overly broad substring match
- Completion doesn't actually complete sub-subcommands (limitation noted inline)
- macOS `bootstrap` runs unconditionally after every install
- `daemonStatus` darwin: `launchctl print` parsing is fragile to format changes
- `parseInstallFlags` silently ignores unknown flags

All cosmetic / future-cleanup material; none gate merge.

Refs #155


IceRhymers:
- [x] Windows
- [x] Linux
- [x] Mac OS

PR #163, opened 2026-05-13
feat(cli): --write-claude-config flag for scriptable settings.json bootstrap

### LINKED ISSUE #162: feat(cli): --write-claude-config flag — scriptable settings.json bootstrap without starting the proxy

IceRhymers:
Closes #162 — Path B reshape as a scriptable top-level flag.

## What

New top-level `--write-claude-config` flag that runs the wrapper's first-run `settings.json` bootstrap and exits. No proxy startup, no port binding, no child process.

```bash
databricks-claude --write-claude-config
databricks-claude --write-claude-config --profile databricks-ai-inference
databricks-claude --write-claude-config --port 49153 --otel-metrics-table main.foo.bar
```

Designed for MDM / fleet init scripts, and as a cleaner alternative to the `databricks-claude --headless` + Ctrl+C workaround (which required binding the proxy port even when the user only wanted the config write — a problem when a daemon already owns that port).

## Behaviour

1. Parse flags; new early-exit branch placed AFTER global wiring (`cli.SetMDMReader`) and the argv[0] credential-helper dispatch, alongside the existing `--print-env` / `--install-hooks` / `--version` exits.
2. Resolve profile (flag → state → MDM → `"DEFAULT"`).
3. Resolve port; do NOT bind it. Compute `proxyURL`.
4. `DiscoverHost(profile, "")` — on failure, `log.Fatalf` with the existing actionable `Run 'databricks auth login --profile %s' first` message. **No settings.json write happens on auth failure.**
5. Build `wcOtelEnv` from explicit `--otel-*-table` flags merged with the canonical `databricksFullSetupEnv()` keys.
6. Call `bootstrapSettings` (same path the normal startup uses).
7. Print `databricks-claude: wrote env block to ~/.claude/settings.json (profile=<P>, base_url=<U>)` to stderr, exit 0.

Idempotent — `ensureConfig` already short-circuits when the env block matches.

## Pipeline

Implemented via OMC: ralplan skipped (issue #162 body was plan-quality with explicit ACs) → autopilot (37 turns, 4.9 min, $1.42) → **two rounds of cross-lane review**.

### Round 1: REQUEST_CHANGES (1 MAJOR — test tautology)

The integration test `TestWriteClaudeConfig_BootstrapWritesFullEnvBlock` built the six-key `otelEnv` inline at the test site and called `bootstrapSettings` directly. Classic "AC-positive vs test-negative" gap: a regression deleting any of the four `ANTHROPIC_DEFAULT_*_MODEL` lines from `main.go` would not fail CI because the test's literal duplicated the implementation's literal.

Fix: extracted `databricksFullSetupEnv() map[string]string` — single source of truth used by BOTH:
- The new `--write-claude-config` branch in `main.go`
- The normal-startup `needsFullSetup` block in `main.go`
- Both test sites (`TestWriteClaudeConfig_BootstrapWritesFullEnvBlock` + new `TestDatabricksFullSetupEnv_KeyCoverage`)

The coverage test pins the EXACT key set (count + presence + spot-checks on the two most drift-prone values).

Bidirectional check performed: temporarily deleted `ANTHROPIC_DEFAULT_SONNET_MODEL` from the helper; the coverage test failed cleanly with "returned 5 keys, want exactly 6" + "missing required key ANTHROPIC_DEFAULT_SONNET_MODEL". Restored.

Round-1 also flagged some MINORs/NITs (`wcOtelEnv` naming smell, triple `loadState()` calls in the new branch, no auth-failure regression test). These mirror pre-existing patterns; deferred per round-1 reviewer's own NIT classification.

### Round 2: APPROVE

Round-2 reviewer ran their own independent bidirectional check (deleted `ANTHROPIC_DEFAULT_HAIKU_MODEL`, watched both `TestDatabricksFullSetupEnv_KeyCoverage` fire with explicit messages, restored). Verified:
- Helper extraction is genuinely single-source-of-truth (`grep ANTHROPIC_DEFAULT_OPUS_MODEL main.go` → one hit, the helper definition).
- Normal-startup path is behaviorally byte-equivalent to pre-amend.
- No name-collision risk between OTEL keys and helper keys.
- All quality gates pass (`go test ./... && go vet ./... && gofmt -l . && GOOS={darwin,windows,linux} go build ./...`).

## Acceptance criteria from #162

- [x] `--write-claude-config` writes the full bootstrap env block and exits 0 without binding port or launching `claude`
- [x] `--profile` / `--port` honoured when paired
- [x] Auth-failure path: `DiscoverHost` failure → `log.Fatalf` with the actionable error, no half-baked write (call sequence verified — bootstrap happens strictly after `DiscoverHost` succeeds)
- [x] OTEL flags honoured and persist via state (same machinery as normal startup)
- [x] `handleHelp` documents the flag
- [x] Completion offers `--write-claude-config`
- [x] README "Using `serve` with Claude Code" section recommends the new flag, with `--headless` workaround retained as fallback
- [x] Test coverage: parser test + integration test asserting against the canonical helper + new `TestDatabricksFullSetupEnv_KeyCoverage` coverage guard

## Non-blocking nits deferred

- `wcOtelEnv` naming carries non-OTEL keys (model routing, custom headers) — mirrors pre-existing `otelEnv` idiom; renaming both sites would be a separate refactor.
- `loadState()` called three times in the new branch — pre-existing pattern in adjacent code, no behavior impact.
- No dedicated test for "no settings.json write on `DiscoverHost` failure" — the call sequence is correct by construction and the auth path is exercised elsewhere.

## Out of scope

- `--uninstall` / `--clear` counterpart (not asked for in #162).
- Auto-detecting a running daemon's port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,9 +517,7 @@ Once you've installed the daemon (`serve install`), Claude Code still needs to k
 ```bash
 # 1. One-time: bootstrap settings.json with the right env block
 #    (proxy URL, fake auth token, Databricks model routing, custom headers).
-databricks-claude --headless
-#    Wait for: PROXY_URL=http://127.0.0.1:49153
-#    Then stop it (Ctrl+C).
+databricks-claude --write-claude-config
 
 # 2. Start the daemon as a service (or run `serve` directly).
 databricks-claude serve install
@@ -527,16 +525,29 @@ databricks-claude serve install
 
 `claude` will now route through whichever instance is listening on `127.0.0.1:49153` — the daemon, in this case.
 
-**Why bootstrap via `--headless` instead of hand-editing `settings.json`?**
+Optional: pair with `--profile`, `--port`, or `--with-websearch` if you use a non-default workspace or want local web-search/web-fetch fulfillment:
+
+```bash
+databricks-claude --write-claude-config --profile my-workspace --port 49153
+
+# Enable local web_search/web_fetch fulfillment in the daemon:
+databricks-claude --write-claude-config --with-websearch
+```
+
+`--with-websearch` (and its sibling `--websearch-backend`/`--websearch-fetch-budget` knobs) persists to `~/.claude/.databricks-claude.json` so the daemon picks it up on its next start. Without this, the daemon serves stock Anthropic web-tool requests, which Databricks FMAPI does not currently support — `claude` would then fail web searches even though the proxy is otherwise healthy.
+
+**Why bootstrap via `--write-claude-config` instead of hand-editing `settings.json`?**
 
 The wrapper writes more than `ANTHROPIC_BASE_URL` on first run. It also writes Databricks-specific model routing (`ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, `ANTHROPIC_DEFAULT_HAIKU_MODEL`), the `x-databricks-use-coding-agent-mode` custom header, and the experimental-betas flag — and the model names are versioned (`databricks-claude-opus-4-7`, etc.), so they change over time. Letting the wrapper write them keeps you in sync with whatever the current shipping binary thinks is correct. Hand-edits get stale.
 
-These env keys persist after `--headless` exits — they're written via the bootstrap path (`ensureConfig`), not the per-session lifecycle (which restores). So one `--headless` run is genuinely one-time setup.
+The write is idempotent — `ensureConfig` short-circuits when the env block already matches.
+
+**Fallback (if `--write-claude-config` is unavailable):** Use `databricks-claude --headless`, wait for `PROXY_URL=http://127.0.0.1:49153`, then stop it (Ctrl+C). This works but binds the proxy port unnecessarily — use `--write-claude-config` instead.
 
 **Notes:**
 
-- **The daemon does NOT mutate `~/.claude/settings.json`.** That's the whole point of the daemon vs. the per-session CLI wrapper — it lives outside the per-tool lifecycle. The one-time `--headless` bootstrap above is the wrapper doing its first-run setup; subsequent daemon restarts do not touch your settings.
-- **Re-bootstrap when model names drift.** If you upgrade `databricks-claude` and the project ships new default model names, re-run `databricks-claude --headless` once to refresh them. The bootstrap is idempotent and only writes keys that differ.
+- **The daemon does NOT mutate `~/.claude/settings.json`.** That's the whole point of the daemon vs. the per-session CLI wrapper — it lives outside the per-tool lifecycle. The one-time `--write-claude-config` bootstrap above is the wrapper doing its first-run setup; subsequent daemon restarts do not touch your settings.
+- **Re-bootstrap when model names drift.** If you upgrade `databricks-claude` and the project ships new default model names, re-run `databricks-claude --write-claude-config` once to refresh them. The bootstrap is idempotent and only writes keys that differ.
 - **OTEL tables persist to state.** Run `databricks-claude serve --otel-metrics-table foo --otel-logs-table bar` once; the daemon (or its installed service) picks them up from `~/.claude/.databricks-claude.json` on every restart thereafter.
 - **Don't run the CLI wrapper (`databricks-claude claude …`) at the same time as the daemon for the same workspace.** Pick one deployment mode per workspace; mixing both means two proxies fighting over the same port and settings block.
 - **Hooks coexist cleanly.** If you've also installed SessionStart hooks (`databricks-claude --install-hooks`), they probe the daemon's `/health` and no-op when it's running, falling back to per-session proxy only if the daemon is down.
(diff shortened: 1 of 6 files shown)


IceRhymers:
When using this flag --with-websearch doesn't initialize the config properly, meaning when running the deamon and running a standard `claude` it won't do a webserach. This flag must also be exposed and perform correctly.

IceRhymers:
Fixed in [`8da1957`](https://github.com/IceRhymers/databricks-claude/commit/8da1957). The `--write-claude-config` branch now resolves and persists `--with-websearch` / `--websearch-backend` / `--websearch-fetch-budget` to `~/.claude/.databricks-claude.json` so the daemon picks them up on its next start.

Worth flagging: websearch is a **proxy-side** feature controlled entirely by state — it doesn't add any key to the settings.json env block. So the integration is one-directional: `--write-claude-config --with-websearch` writes state.WithWebSearch=true, then `serve` reads it on startup. settings.json is untouched by websearch (which is correct — websearch belongs to the proxy, not the client).

Confirmation-line output now includes `websearch=true|false` so a scripted invocation has visible feedback:

```
$ databricks-claude --write-claude-config --with-websearch
databricks-claude: wrote env block to ~/.claude/settings.json (profile=DEFAULT, base_url=http://127.0.0.1:49153, websearch=true)
```

Tests added:
- `TestParseArgs_WriteClaudeConfig_WithWebSearch` — verifies the parser sets the `*Set` sentinels alongside `WriteClaudeConfig=true` for the three-flag combo
- `TestWriteClaudeConfig_PersistsWebSearchState` — drives the resolution+persistence block end-to-end against a temp state file; if a future regression drops one of the three writes, the per-field assertion fails

README updated under "Using `serve` with Claude Code" with a worked example.

All quality gates still green (`go test ./...`, `go vet`, `gofmt`, cross-OS builds for darwin/windows/linux).

PR #193, opened 2026-06-03
chore: release 1.2.0

### LINKED ISSUE #191: feat(proxy): rewrite mismatched item_ids in Responses API SSE streams
## Overview

Add a Responses-API SSE rewriter to `pkg/proxy` so OpenAI-compatible clients (opencode via `@ai-sdk/openai`) don't break on Databricks AI Gateway streams. The gateway re-encodes Responses-API SSE and emits a different id in `response.output_item.added` (`item.id`) than in the downstream `response.output_text.*` / `response.content_part.*` events (`item_id`), tripping the AI SDK stream parser with `text part <id> not found`.

This is the Go-side fix for [IceRhymers/opencode#1](https://github.com/IceRhymers/opencode/issues/1). The proxy already sits in the byte path between opencode and the gateway, so we fix it once here in shared `pkg/proxy` instead of carrying a fork delta in opencode's TypeScript `@ai-sdk` integration.

## Design (mirror the existing websearch passive-gate pattern)

- **New `pkg/proxy/responses_rewriter.go`:** `pumpResponsesSSE(ctx, w, src, verbose)`. Reuse `splitSSEFrames` / `parseSSEFrame` from `sse_rewriter.go`. Per-stream `map[int]string` caching `output_index -> item.id` from `output_item.added` (message items only — reasoning/tool_call items intentionally ignored). Delete the entry on `output_item.done` so reused indices don't carry a stale id. Rewrite `item_id` on downstream events when it disagrees with the cached canonical id. `log.Printf` once per `(output_index, originalId)` pair for observability.
- **New `Config.ResponsesRewrite ResponsesRewriteSettings{ Enabled bool }`** field, sibling to `WebSearch`. Documented identically: when `Enabled=false`, inference requests are forwarded byte-identically and the struct has no effect.
- **Wire the gate inside the `!rewroteTools.Any()` fast path** in `inferenceHandler` (`websearch_handler.go` ~line 128): if `config.ResponsesRewrite.Enabled && isResponsesPath(r.URL.Path) && isSSEResponse(resp)` -> strip Content-Length, run `pumpResponsesSSE`; otherwise fall through to the existing `flushingCopy`. **Must NOT depend on `ws.Enabled`** — opencode runs with websearch off, so the new gate has to live in the byte-identity fast path, not the websearch rewrite branch.
- **New helper `isResponsesPath(p)`:** path contains `/responses`.
- **Passthrough verbatim** on: parse failure, unknown event types, `[DONE]`, reordered events (delta before added), and any event that doesn't need a rewrite (byte-equivalent for conforming streams).

## Why this seam

`NewServer` always registers `inferenceHandler` on the catch-all `/`. opencode's `/responses` requests aren't `/v1beta` or `/otel`, so they hit it unconditionally. With `WebSearch.Enabled=false` (the opencode/codex case) the request is never rewritten, `rewroteTools` stays zero, and the handler takes the `!rewroteTools.Any()` byte-identity fast path. That fast path is the only place the Responses rewriter can branch in for opencode — hence the gate must be independent of `ws.Enabled`. Claude Code never emits `/responses`, so it stays byte-identical.

## Acceptance Criteria

- [ ] `pumpResponsesSSE` rewrites mismatched `item_id` -> canonical `item.id`
- [ ] Reused `output_index` after `output_item.done` does not carry a stale id
- [ ] Reordered (delta-before-added) events pass through unchanged
- [ ] Conforming stream is byte-equivalent (regression test, mirrors `TestProxy_WebSearchDisabled_ByteIdenticalForward`)
- [ ] Content-Length stripped on the rewrite path
- [ ] `ResponsesRewrite.Enabled=false` -> byte-identical forward (Claude Code untouched)
- [ ] `responses_rewriter_test.go` covers all the above
- [ ] `go test ./...` and `go vet ./...` clean; CI green
- [ ] release-please `feat:` -> minor bump (v1.1.0 -> v1.2.0)

## Consumers

databricks-opencode bumps to the resulting v1.2.0 and enables the gate (tracked separately). databricks-codex also shares `pkg/proxy`, so the rewriter is available there if its SSE path ever needs it.

IceRhymers:
:robot: I have created a release *beep* *boop*
---

### Features

* **proxy:** rewrite mismatched item_ids in Responses API SSE streams ([250dba4](https://github.com/IceRhymers/databricks-claude/commit/250dba4977d351a6863f76b936d8687345d299c6)), closes [#191](https://github.com/IceRhymers/databricks-claude/issues/191)

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).


IceRhymers:
🤖 Created releases:

- [v1.2.0](https://github.com/IceRhymers/databricks-claude/releases/tag/v1.2.0)

:sunflower:

PR #231, opened 2026-07-18
fix(codex): register proxy as top-level default provider, not a profile (#230)

### LINKED ISSUE #230: databricks-codex: config.toml profile patch breaks on Codex 0.134.0+ (legacy `profile = "..."` no longer supported)
## Summary

`databricks-codex` no longer works with recent versions of the OpenAI Codex CLI (0.134.0+). Codex removed support for the legacy top-level `profile = "..."` selector and for reading `[profiles.<name>]` tables out of `config.toml` via `--profile`. Our `config.toml` patcher writes exactly the config shape Codex now rejects, so Codex refuses to start:

```
Error loading configuration: legacy `profile = "databricks-proxy"` config is no longer supported;
use `--profile databricks-proxy` with `databricks-proxy.config.toml` instead
```

This is a hard startup failure (Codex exits before doing anything), so any user on an up-to-date Codex CLI is fully blocked.

## Repro

1. Install a current Codex CLI (`>= 0.134.0`).
2. Run `databricks-codex` (or `databricks-agents codex`).
3. The patcher writes `profile = "databricks-proxy"` + `[profiles.databricks-proxy]` into `~/.codex/config.toml`, and Codex aborts with the error above.

## Root cause

`internal/codex/tomlconfig/tomlconfig.go` patches `~/.codex/config.toml` using the pre-0.134 profile model:

- `managedRootKeys = []string{"profile"}` (`tomlconfig.go:86`) and `patchRootKey(content, "profile", "\"databricks-proxy\"")` (`tomlconfig.go:107`) write the top-level `profile = "databricks-proxy"` selector — **now rejected outright**.
- `buildProfileSection` writes `[profiles.databricks-proxy]` (`model_provider`, `model`) into `config.toml` (`tomlconfig.go:110-111,136-172`). Even with the root `profile` key removed, `--profile databricks-proxy` **no longer reads `[profiles.*]` from `config.toml`** on new Codex, so the provider/model wiring would silently not apply.

## What Codex changed (0.134.0+)

- The top-level `profile = "<name>"` selector in `config.toml` is unsupported and errors.
- `--profile <name>` no longer consumes `[profiles.<name>]` from `config.toml`. Instead Codex loads `~/.codex/config.toml`, then **overlays** `~/.codex/<name>.config.toml`.
- Migration per upstream: move per-profile settings into `~/.codex/<name>.config.toml`, and remove the `[profiles.<name>]` table + `profile = "<name>"` selector from `config.toml`.

Refs: openai/codex#24076 (stop consuming legacy config profiles), openai/codex#24858 (legacy profile config errors instead of migrating/warning).

### Option A (recommended): configure the proxy as the top-level default provider — no profiles at all

Codex still supports selecting a **default provider** via the top-level `model_provider` key (unaffected by the profile deprecation — see [config reference](https://developers.openai.com/codex/config-reference)). We can drop the profile mechanism entirely and patch `config.toml` to:

```toml
model = "databricks-<...>"
model_provider = "databricks-proxy"

[model_providers.databricks-proxy]
name = "Databricks Proxy"
base_url = "http://127.0.0.1:<port>"
api_key = "databricks-proxy"
wire_api = "responses"
```

Concretely:

1. **Stop writing the root `profile` key** — drop `"profile"` from `managedRootKeys` and the `patchRootKey` call.
2. **Replace `[profiles.databricks-proxy]` with top-level `model_provider` + `model`** — write `model_provider = "databricks-proxy"` and `model = "..."` as managed root keys instead of a profile table. Keep `[model_providers.databricks-proxy]` and `[otel]` as-is.
3. **No `--profile` injection needed** — plain `codex` picks up the default provider directly.

**Why this is preferred:** it makes bare `codex` invocations route through the proxy with **no argv injection**. That's essential for the flows where we *can't* inject a flag — the **hooks path** (`[features] hooks = true` in `config.toml`; the user runs plain `codex`, which fires the SessionStart hook that spawns `databricks-codex serve`) and any daemon/detached spawn. Option B below can't cover those because the profile only activates when `--profile` is on the command line.

The trade-off: `model_provider` is a *global* default — we'd be pointing the user's default provider at the proxy while wrapped (the current restore-on-exit discipline already handles putting it back). Confirm this is acceptable vs. the scoping a profile gave.

### Option B (fallback): per-profile overlay file + `--profile` injection

If we want to keep proxy config scoped to a named profile rather than flipping the global default:

1. Stop writing the root `profile` key.
2. Move the profile body (`model_provider`, `model`) into a wrapper-owned `~/.codex/databricks-proxy.config.toml` overlay; keep `[model_providers.databricks-proxy]`/`[otel]` in `config.toml`.
3. Inject `--profile databricks-proxy` into the Codex child argv (wrapper mode via `core.Run`/`CodexArgs` **and** `serve`), unless the user passed their own `--profile`.

**Limitation:** does **not** fix the hooks/daemon-driven plain-`codex` path (no `--profile` there). If we go this route we'd need Option A's default-provider config *in addition* for those flows — which is why Option A alone is simpler and more complete.

### Open questions / risks

- **Version compat.** Older Codex (`< 0.134.0`) still accepts the old `profile`/`[profiles.*]` shape. Option A's top-level `model_provider` works on both old and new Codex, so a clean cut to Option A likely avoids any version branching — worth verifying against the oldest Codex we intend to support.
- **Global-default scoping (Option A).** Setting `model_provider` at root changes the user's default provider for the duration of the wrapped session. The existing surgical patch/restore (or codex's patch-every-session model) must put the prior default back cleanly on exit.
- **Restore semantics.** Confirm the patcher's save-originals/restore logic correctly captures and restores a pre-existing root `model_provider`/`model` (sentinel when absent), same discipline the current `[profiles.*]` patch uses.
- Update `internal/codex/tomlconfig/AGENTS.md`, `CLAUDE.md` (codex config-patch description), and tests to match.

## Integration

Folds into the unify epic (#196) work for `databricks-codex`. Should land on the `feat/unity-agent-tooling` integration branch.

IceRhymers:
Fixes #230. Targets the **`feat/unity-agent-tooling`** integration branch (per the issue's Integration note).

## Problem
Codex 0.134+ removed the legacy profile model: a root `profile = "..."` selector in `~/.codex/config.toml` is now a **hard startup error**, and `--profile <name>` no longer reads `[profiles.<name>]`. `databricks-codex` patched config.toml with exactly that shape, so Codex refused to start:

```
Error loading configuration: legacy `profile = "databricks-proxy"` config is no longer supported;
use `--profile databricks-proxy` with `databricks-proxy.config.toml` instead
```

## Fix (Option A — top-level default provider)
Patch the proxy as the top-level **default provider** instead of a named profile:
```toml
model = "databricks-<...>"
model_provider = "databricks-proxy"

[model_providers.databricks-proxy]
name = "Databricks Proxy"
base_url = "http://127.0.0.1:<port>"
api_key = "databricks-proxy"
wire_api = "responses"
```
Bare `codex` then routes through the proxy with **no argv injection** — essential for the hooks/daemon path (plain `codex`, no `--profile` injection point). Top-level provider keys predate 0.134, so this works on old and new Codex with no version detection.

## One-directional migration
codex is patch-and-leave (`Backup` never called at runtime, `Restore` a no-op), so returning users carry the fatal shape persisted. `Patch` now strips it:
- `removeLegacyRootProfile` deletes a root `profile = "databricks-proxy"` line (exact value match; **not** recorded for restore, so it can't be resurrected).
- `removeSection` strips the stale `[profiles.databricks-proxy]` block.
- A user's **own** non-proxy root `profile = "<other>"` is left untouched, with a non-fatal warning.

## Notable correctness detail
Model resolution now reads the **on-disk content** (not `m.original`, which is nil at runtime) and keeps all three precedence branches: `--model` > preserve existing root `model` > resolved default. The consensus review caught that the old preserve-if-present logic read `m.original` and was therefore dead in production.

## Tests
- Golden + unit tests rewritten to the new shape; golden bytes pinned from real output.
- Precise negative assertions (scoped to `[profiles.databricks-proxy]` / root `profile = "databricks-proxy"`, never blanket).
- New: `TestPatch_NeverEmitsLegacyProfile`, `TestPatch_MigratesLegacyProfileShape` (idempotent), `TestPatch_ModelPrecedenceRuntime` (no `Backup()` — runtime-representative), `TestPatch_MigrationNonDestructiveToForeignProfile`.
- Wrapper-mode ≡ serve-mode byte-parity preserved (`profile_codex.go`/`serve_codex.go`/`PatchConfig`/`PatchRequest` unchanged).

## Verification
- `go test ./internal/codex/tomlconfig/... ./cmd/databricks-codex/...` — all pass
- `go vet ./...` clean; `go build ./cmd/databricks-codex` and `make build` green
- Codex 0.134+ runtime smoke (fresh launch, hooks bare-`codex`, upgrade path) is a **manual** step — a real Codex binary isn't available in CI here; unit/golden tests prove what we write, not that Codex accepts it.

## Docs
`internal/codex/tomlconfig/AGENTS.md`, `README.md`, `CLAUDE.md` updated to the default-provider shape + migration rationale.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/codex/tomlconfig/tomlconfig.go
```diff
@@ -103,12 +100,47 @@ func (m *Manager) Patch(cfg PatchConfig) error {
...
+
+	// --- Model resolution ---
+	// databricks-codex is patch-and-leave: Backup() is never called on the
+	// runtime path, so m.original is nil in production. Read the on-disk
+	// `content` string, NOT m.original.
+	//
+	// Precedence: an explicit --model always wins; otherwise preserve the
+	// user's existing root model; otherwise fall back to the resolved model
+	// (saved state or built-in default).
+	existingRoot := m.findRootModel(content)
```
IceRhymers, line 126: **LOW — migration is lossy for a model that lived *only* inside `[profiles.databricks-proxy]`.**

`removeSection("profiles.databricks-proxy")` runs before this line, so `findRootModel(content)` cannot see a `model` the user had hand-set *inside* the old proxy profile section — resolution falls through to `cfg.Model` (state/default). For the common cases this is a no-op (the value was written from state, or a root `model` was already present and is preserved). The only real loss is a user who manually edited the model inside the legacy proxy profile section to a value not reflected in `~/.codex/.databricks-codex.json`.

Consciously accepted as out-of-scope for #230 (rare manual-edit edge; reintroducing the deleted `findModelInSection` to cover it would add back the complexity this PR removed). Flagging for the record.

internal/codex/tomlconfig/tomlconfig.go
```diff
@@ -131,44 +163,41 @@ func (m *Manager) Patch(cfg PatchConfig) error {
...
+		trimmed := strings.TrimSpace(line)
+		if !isRootKey(trimmed, "profile") || inAnySection(lines, i) {
+			continue
+		}
+		parts := strings.SplitN(trimmed, "=", 2)
+		if len(parts) != 2 {
+			continue
+		}
+		val := strings.Trim(strings.TrimSpace(parts[1]), `"`)
+		if val != "databricks-proxy" {
```
IceRhymers, line 187: **NIT — a legacy `profile = "databricks-proxy"  # comment"` line is treated as foreign and not migrated.**

`strings.Trim(strings.TrimSpace(parts[1]), "\"")` on `"databricks-proxy" # note` trims only the leading quote, yielding `databricks-proxy" # note`, which fails the exact-match and hits the foreign-profile warning branch instead of being stripped. Extremely unlikely for a tool-authored line (we never emit trailing comments), so not worth the added parsing. Noting for completeness.

internal/codex/tomlconfig/tomlconfig.go
```diff
@@ -83,11 +81,10 @@ func (m *Manager) Backup() error {
 }
 
 // managedRootKeys lists top-level keys we manage.
-var managedRootKeys = []string{"profile"}
+var managedRootKeys = []string{"model", "model_provider"}
```
IceRhymers, line 84: **NIT (pre-existing) — `managedRootKeys` and `managedSections` are declared but never referenced.**

`Patch` calls `patchRootKey`/`patchSection` with literals; these two package vars are documentation-only and have no readers (confirmed by grep). Not introduced by this PR — it only updates their values — but since the diff touches them: consider either wiring them into `Patch`/`Restore` as the source of truth or dropping them to avoid drift between the list and the actual managed keys. Non-blocking.

